PR #34, opened 2019-08-26
remove quotation marks of column names in sql server queries

chandryou:
I think errors occur because of quotation marks of column names in sql server queries. So I removed it. 

(only '@field_name' will work, but '@cdmDatabase.@field_name' doesn't work)


chandryou:
@clairblacketer Sorry, I misunderstood the original code. You're right. 
I'll post an error to the issue, @alondhe

PR #72, opened 2019-11-26
Run DQD for a cohort

clairblacketer:
This PR includes the start of a vignette but it is not complete. Additionally we need to think about best practices to identify the cohort in the JSON file.


R/execution.R
```diff
@@ -133,6 +40,9 @@ executeDqChecks <- function(connectionDetails,
                             cdmDatabaseSchema,
                             resultsDatabaseSchema,
                             cdmSourceName,
+                            cohort = FALSE,
```
alondhe: Can you add these to the meta tagging of the function?

alondhe: Also, do we need the cohort parameter? Could the presence of cohortDefinitionIds indicate that we want to run this for a cohort or cohorts?

clairblacketer:
Closing - this was added in other PR

Issue #65, opened 2019-10-30
Field level CSV has unprintable characters.


don-torok:
Note_NLP.term_temporal has unprintable characters in the 'etlConventions' column.

Issue #86, opened 2020-02-11
add dqd version to metadata


fdefalco:
Allow metadata to include the current version of this package used to generate the output.

alondhe:
In the develop branch of my fork, I've added this to the metadata section

alondhe:
Fixed in latest commit

jimfrankfort:
thanks for updating the DQD! We ran this and see the version listed as DQD 1.0.0. This is nice but not sufficient to address our use case. We have 3 data partners and we want to all run the exact same DQD. I don't think 1.0.0 does it. the prior release had some bugs fixed (e.g. case statement in field names)....what would that version be? Although we can sync our DQDs by loading a specific commitID ID, it would be nice to be able to tell from the meta data. Apologies in advance if I am mistaken. thx

Issue #81, opened 2020-02-03
Incorrect quotes in CAST usage in PostgreSQL


SulevR:
Tried to run it on PostgreSQL database, but most of the SQL scripts end up with error like **'ERROR: column "CAST(birth_datetime AS DATE)" does not exist'.**

It seems that the produced SQL is not correct - uses quotes where it should not:
`...select case when count("CAST(birth_datetime AS DATE)") = 0 ...`

I tested that the same script without quotes works well:
`...select case when count(CAST(birth_datetime AS DATE)) = 0 ...`

Full stacktrace here:

```
DBMS:
postgresql

Error:
org.postgresql.util.PSQLException: ERROR: column "CAST(birth_datetime AS DATE)" does not exist
  Position: 219

SQL:

/*********
FIELD LEVEL check:
CDM_FIELD - verify the field exists

Parameters used in this template:
cdmDatabaseSchema = ohdsi
cdmTableName = PERSON
cdmFieldName = CAST(birth_datetime AS DATE)

**********/


SELECT num_violated_rows, CASE WHEN denominator.num_rows = 0 THEN 0 ELSE 1.0*num_violated_rows/denominator.num_rows END AS pct_violated_rows
FROM
(
  select num_violated_rows from
  (
    select 
      case when count("CAST(birth_datetime AS DATE)") = 0 then 0
      else 0
    end as num_violated_rows
    from ohdsi.PERSON
  ) violated_rows
) violated_row_count,
( 
	SELECT 1 as num_rows
) denominator
;

R version:
R version 3.6.2 (2019-12-12)

Platform:
x86_64-apple-darwin15.6.0

Attached base packages:
- stats
- graphics
- grDevices
- utils
- datasets
- methods
- base

Other attached packages:
- magrittr (1.5)
- snow (0.4-3)
```

deebowden:
This is occurring in Sql Server as well.
SQL Server : Microsoft SQL Server 2016 (SP2-CU10) (KB4524334) - 13.0.5492.2 (X64) 
	Oct  4 2019 19:14:08 
	Copyright (c) Microsoft Corporation
	Standard Edition (64-bit) on Windows Server 2016 Datacenter 10.0 <X64> (Build 14393: ) (Hypervisor)

R packages: see bottom


with the following fields:
CDM_TABLE_NAME	CDM_FIELD_NAME
CONDITION_ERA	CAST(condition_era_start_date AS DATE)
[...]

Issue #80, opened 2020-01-17
Add FIELD name to the UI


alondhe:
In the current table UI, we don't have the field name available except for in the description. It would be easier to read if we include it.

![image](https://user-images.githubusercontent.com/8037742/72644588-39d4a880-3926-11ea-8bab-bf925f47e78d.png)

jimfrankfort:
Until a persistent unique identifier is created for data quality checks, I needed to create one to combine DQD results from multiple CDM instances.  It would be great if you would include conceptID, unitConceptID, and quality check denominator count in addition to CDM_Field_Name in the JSON output.  I need the denominator to combine results at the DQ check level (sum of numerators, sum of denominators, recalc %).
  Fields needed to uniquely ID a DQ check are:
Check_Level, Check_Name, CDM_Table_Name, CDM_Field_Name, conceptID, and unitConceptID, 
We are string processing the query text to get these field values.....but it would be nice if it were part of the JSON output.  Also, we are estimating the denominator using numerator and % violated....but that doesn't work when there are 0 failures :-).....need that output too.

thanks for considering this!.....jimf

alondhe:
@jimfrankfort added the denominator for now in my fork. I'll look into adding the other fields you mentioned.

PR #88, opened 2020-02-19
UI enhancements and SQL fixes

alondhe:
Fixes:
#65
#81
#79
#93

Enhancements:
#86
#80
#77


inst/sql/sql_server/concept_plausible_gender.sql
```diff
@@ -13,18 +13,19 @@ plausibleGender = @plausibleGender
...
 	SELECT COUNT_BIG(*) AS num_violated_rows
 	FROM
 	(
-		SELECT @cdmTableName.* 
-		FROM @cdmDatabaseSchema.@cdmTableName
-			INNER JOIN @cdmDatabaseSchema.person
-			ON @cdmTableName.person_id = person.person_id
-		WHERE @cdmFieldName = @conceptId
-		AND person.gender_concept_id <> {@plausibleGender == 'Male'} ? {8507} : {8532} 
+		SELECT A.*
```
fdefalco: Would recommend aliases that were more descriptive than A/B.

inst/sql/sql_server/field_cdm_field.sql
```diff
@@ -7,17 +7,21 @@ Parameters used in this template:
 cdmDatabaseSchema = @cdmDatabaseSchema
 cdmTableName = @cdmTableName
 cdmFieldName = @cdmFieldName
+isPostgresql = @isPostgresql
```
fdefalco: Is this not handled by SqlRender? Unclear as why there is DB specific instruction.  Might just need to understand.

inst/sql/sql_server/concept_plausible_gender.sql
alondhe: Latest commit uses more descriptive aliases

inst/sql/sql_server/field_cdm_field.sql
alondhe: Misunderstanding on my part -- it's not that PG was an issue, it was the casts added in the CSV. I've removed this hack.

alondhe:
@fdefalco and @clairblacketer -- can you review now? I've updated the aliases to be more descriptive, removed the isPostgresql hack, and removed the cast as datetimes from the field level CSV and simply placed that into the relevant SQL templates.

alondhe:
Thanks @fdefalco for the review!

PR #172, opened 2020-08-19
Remove dependency on cdm_source table

### LINKED ISSUE #171: Remove dependency on cdm_source
If the `cdm_source` table is empty, the DQD gives the following error:
```
Error in `$<-.data.frame`(`*tmp*`, "DQD_VERSION", value = "1.0.0") : 
  replacement has 1 row, data has 0
```
Can we make it optional to read from the `cdm_source` table? If there is no record, we just leave the metadata for the DQD empty (or use the name provided in `cdmSourceName`).

MaximMoinat:
Fixes #171

### DIFF
--- a/R/execution.R
+++ b/R/execution.R
@@ -514,7 +514,28 @@ if (conceptCheckThresholdLoc == "default"){
   sql <- SqlRender::render(sql = "select * from @cdmDatabaseSchema.cdm_source;",
                            cdmDatabaseSchema = cdmDatabaseSchema)
   sql <- SqlRender::translate(sql = sql, targetDialect = connectionDetails$dbms)
-  metadata <- DatabaseConnector::querySql(connection = connection, sql = sql)
+  metadata <- tryCatch(
+    expr = {
+      DatabaseConnector::querySql(connection = connection, sql = sql)
+    },
+    error = function(e) {
+      ParallelLogger::logWarn("Retrieving metadata from cdm_source failed.")
+      return(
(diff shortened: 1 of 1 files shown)


vojtechhuser:
+1

R/execution.R
```diff
@@ -514,7 +514,28 @@ if (conceptCheckThresholdLoc == "default"){
   sql <- SqlRender::render(sql = "select * from @cdmDatabaseSchema.cdm_source;",
                            cdmDatabaseSchema = cdmDatabaseSchema)
   sql <- SqlRender::translate(sql = sql, targetDialect = connectionDetails$dbms)
-  metadata <- DatabaseConnector::querySql(connection = connection, sql = sql)
+  metadata <- tryCatch(
+    expr = {
+      DatabaseConnector::querySql(connection = connection, sql = sql)
+    },
+    error = function(e) {
+      ParallelLogger::logWarn("Retrieving metadata from cdm_source failed.")
```
MaximMoinat, line 522: @clairblacketer We could change this into something like 
> "CDM_SOURCE not found, your DQD result will not contain metadata on your run. Are you sure you want to continue?". 

And then you have to manually 'click away' the warning.

clairblacketer:
@Maxim I like that - I agree that it shouldn't stop the run of the DQD as
it is super annoying to have to re-run but that metadata should be included.

fdefalco:
closing in favor of graceful-degradation branch

Issue #162, opened 2020-08-07
cdmField check fails in Oracle 


kirenotneb:
The cdmField check fails in Oracle due to the double quotes present in the query.
 
```
[Level: FIELD] [Check: cdmField] [CDM Table: CARE_SITE] [CDM Field: care_site_name] Error executing SQL: java.sql.SQLSyntaxErrorException: ORA-00904: "care_site_name": invalid identifier  An error report has been created at  output/RDW OMOP/errors/FIELD_cdmField_CARE_SITE_care_site_name.txt
```

The query that DQD executes:
 
```
SELECT num_violated_rows, CASE WHEN denominator.num_rows = 0 THEN 0 ELSE 1.0*num_violated_rows/denominator.num_rows END AS pct_violated_rows,
  denominator.num_rows as num_denominator_rows
FROM (SELECT num_violated_rows FROM (SELECT  CASE when  COUNT("care_site_name") = 0 then 0
      else 0
     END as num_violated_rows
    FROM RDW_OMOP.CARE_SITE cdmTable
    GROUP BY 1) violated_rows
) violated_row_count,
(SELECT 1 as num_rows
FROM DUAL) denominator
;
```

The innermost query is:
```
SELECT  CASE when  COUNT("care_site_name") = 0 then 0
      else 0
     END as num_violated_rows
    FROM RDW_OMOP.CARE_SITE cdmTable
    GROUP BY 1;
```
 
refers to the column name `care_site_name` in lower case, but our database it is capitalized `CARE_SITE_NAME` due to the manner how Oracle interprets the DDL for table creation.

This appears to have also been an issue with PostgreSQL as well (#151 ).

clairblacketer:
@alondhe can you take a look at this for Oracle? I don't have an Oracle environment to test this in.

PR #101, opened 2020-03-08
Oracle and BigQuery fixes

alondhe:
#98 -- dynamic casting of datetimes in queries that use comparisons. This will ensure BigQuery date comparisons will work.
#100 -- use upper case for cdm field names to allow us to use quotes around the field name in field_cdm_field check. This will ensure we can quote the CDM field "offset" and not run into a keyword error on Postgres **and** not run into a case sensitivity issue in Oracle

PR #183, opened 2020-10-19
Double quotes produce errors in Oracle

kirenotneb:
Double quotes are interpreted with an explicit case in Oracle which does not conform with the DDL for v5.3.1 for the OMOP CDM, which results in all conformance checks failing for the `cdmField` checks in Oracle. This PR addresses #162.

### DIFF
--- a/inst/csv/OMOP_CDMv5.2.2_Field_Level.csv
+++ b/inst/csv/OMOP_CDMv5.2.2_Field_Level.csv
@@ -155,7 +155,7 @@ NOTE_NLP,note_nlp_id,Yes,0,,integer,0,,,,Yes,0,,No,,,,,,,,,,,No,,,Yes,0,,No,,,No
 NOTE_NLP,note_id,Yes,0,,integer,0,,,,No,,,Yes,0,,NOTE,NOTE_ID,,,,,,,No,,,Yes,0,,No,,,No,,,No,,,,,,,,,,,,,,,No,,,No
 NOTE_NLP,section_concept_id,No,,,integer,0,,,,No,,,Yes,0,,CONCEPT,CONCEPT_ID,,,,,,,Yes,0,,Yes,100,,No,,,No,,,No,,,,,,,,,,,,,,,No,,,No
 NOTE_NLP,snippet,No,,,varchar(250),0,,,,No,,,No,,,,,,,,,,,No,,,Yes,100,,No,,,No,,,No,,,,,,,,,,,,,,,No,,,No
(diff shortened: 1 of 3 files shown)


inst/csv/OMOP_CDMv5.2.2_Field_Level.csv
```diff
@@ -155,7 +155,7 @@ NOTE_NLP,note_nlp_id,Yes,0,,integer,0,,,,Yes,0,,No,,,,,,,,,,,No,,,Yes,0,,No,,,No
 NOTE_NLP,note_id,Yes,0,,integer,0,,,,No,,,Yes,0,,NOTE,NOTE_ID,,,,,,,No,,,Yes,0,,No,,,No,,,No,,,,,,,,,,,,,,,No,,,No
 NOTE_NLP,section_concept_id,No,,,integer,0,,,,No,,,Yes,0,,CONCEPT,CONCEPT_ID,,,,,,,Yes,0,,Yes,100,,No,,,No,,,No,,,,,,,,,,,,,,,No,,,No
 NOTE_NLP,snippet,No,,,varchar(250),0,,,,No,,,No,,,,,,,,,,,No,,,Yes,100,,No,,,No,,,No,,,,,,,,,,,,,,,No,,,No
-NOTE_NLP,offset,No,,,varchar(50),0,,,,No,,,No,,,,,,,,,,,No,,,Yes,100,,No,,,No,,,No,,,,,,,,,,,,,,,No,,,No
+NOTE_NLP,'offset',No,,,varchar(50),0,,,,No,,,No,,,,,,,,,,,No,,,Yes,100,,No,,,No,,,No,,,,,,,,,,,,,,,No,,,No
```
kirenotneb, line 158: As `offset` is a reserved word it must be escaped for SqlRender before it is run in the CDM.

inst/sql/sql_server/field_cdm_field.sql
```diff
@@ -18,7 +18,7 @@ FROM
   select num_violated_rows from
   (
     select
-      case when count_big("@cdmFieldName") = 0 then 0
+      case when count_big(@cdmFieldName) = 0 then 0
```
kirenotneb, line 21: This allows the RDBMS to  naturally interpret the field casing rather than with an explicit casing.

clairblacketer:
Hi @kirenotneb thank you so much for the pull request. We have to add the single quotes because offset is a reserved word in other dialects and taking them out causes failures elsewhere. However, this is on me because @alondhe has PR #101 that has been approved but not merged. I apologize, I will merge that in and send back to you to test.

clairblacketer:
Closing - the explicit interpretation of the case will not work when employing SqlRender. I tried different approaches to this problem and I ended up just removing offset from being evaluated all together. SqlRender translates everything to all caps and therefore there is no good work around.

PR #185, opened 2020-10-29
adding data quality result indexing function

fdefalco:


fdefalco:
removed in favor of separate indexing package.

PR #221, opened 2021-03-17
Added option to remove checks with zero denominator

clairblacketer:
This PR includes an option in the execution function to remove checks with zero denominators (by default set to TRUE). There is a separate function that allows you to pass a JSON file and it will remove the checks with no data.


clairblacketer:
Change to four status: pass, fail, error, no data where no data has a higher priority than pass but lower than fail

clairblacketer:
I am going to close this as it is handled with Dmitry's addition of IS_APPLICABLE and IS_ERROR

PR #262, opened 2021-11-19
add '...' parameter and default parameters to view function

hollorol:
To allow users to use other port and host in viewDqDashboard(), I added '...' parameter and make display.mode and launch.browser an additional parameter where default options were set. With these litte change the user can use any parameter of shiny::runApp() function.

### DIFF
--- a/R/view.R
+++ b/R/view.R
@@ -35,5 +35,5 @@ viewDqDashboard <- function(jsonPath, launch.browser=NULL, display.mode=NULL, ..
       launch.browser=TRUE
   }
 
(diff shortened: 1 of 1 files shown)


kesadae11 (approved):
We tested it in our OMOP data harmonization project. Everything working properly.
Super feature! we can start the dashboard on a fix port, which is very important for us when we configure the firewall.

kesadae11 (approved):

clairblacketer (approved):
Looks great, I tested and everything works well :)

Issue #251, opened 2021-09-20
Error about "THRESHOLD_VALUE": replacement has 0 rows, data has 1 


schuemie:
I have converted SynPuf to CDM 5.2.2, and ran this code:

```r
executeDqChecks(connectionDetails = connectionDetails, 
                cdmDatabaseSchema = cdmDatabaseSchema, 
                resultsDatabaseSchema = resultsDatabaseSchema,
                cdmSourceName = "SynPuf", 
                cdmVersion = "5.2.2")
```

which eventually produces the error
```
Error in `$<-.data.frame`(`*tmp*`, "THRESHOLD_VALUE", value = integer(0)) : 
  replacement has 0 rows, data has 1
```

The most likely cause it [this line](https://github.com/OHDSI/DataQualityDashboard/blob/master/R/execution.R#L541), which will throw an error if `thresholdValue` is empty. Why `thresholdValue` would empty, I don't know.

Here's the log:

```
2021-09-17 05:23:34	[Main thread]	INFO	DataQualityDashboard	FUN	Processing check description: measurePersonCompleteness
2021-09-17 06:27:38	[Main thread]	INFO	DataQualityDashboard	FUN	Processing check description: cdmField
2021-09-17 07:18:11	[Main thread]	ERROR	DataQualityDashboard	3	[Level: FIELD] [Check: cdmField] [CDM Table: NOTE_NLP] [CDM Field: OFFSET] Error executing SQL: org.postgresql.util.PSQLException: ERROR: syntax error at or near "OFFSET"   Position: 267 An error report has been created at  output/errors/FIELD_cdmField_NOTE_NLP_OFFSET.txt
2021-09-17 07:21:04	[Main thread]	ERROR	DataQualityDashboard	3	[Level: FIELD] [Check: cdmField] [CDM Table: COST] [CDM Field: REVENUE_CODE_SOURCE_VALUE] Error executing SQL: org.postgresql.util.PSQLException: ERROR: column "revenue_code_source_value" does not exist   Hint: Perhaps you meant to reference the column "cdmtable.reveue_code_source_value".   Position: 267 An error report has been created at  output/errors/FIELD_cdmField_COST_REVENUE_CODE_SOURCE_VALUE.txt
2021-09-17 07:28:02	[Main thread]	INFO	DataQualityDashboard	FUN	Processing check description: isRequired
[...]

PR #263, opened 2021-11-23
Fixed an empty threshold causing issues

clairblacketer:
Addressing issue #251 where there was an empty threshold value in the field level file and bugs in the concept level file.


clairblacketer:
@fdefalco or @AnthonyMolinaro can one of you take a look at this one? Unfortunately it looks like maybe an additional column was put at the end so it might mask the actual changes made. Let me know if you want to chat about this

AnthonyMolinaro (approved):
Successfully tested against a v5.2.2 CDM on redshift without error.

PR #270, opened 2021-12-08
Add a missing parameter in an execution subroutine in R code

dleprovost:
`.needsAutoCommit` requires a variable named `connectionDetails` which is not provided by its declaration.
The current PR fixes it with the value created by `executeDqChecks`.

(related to #269 )


clairblacketer (approved):
Looks good, thanks for catching that!

PR #273, opened 2022-01-21
Adds support for V5.4

clairblacketer:
@AnthonyMolinaro or @fdefalco please run this against a v5.4 instance and test that it runs and produces expected results


AnthonyMolinaro (approved):
Successfully tested against a v5.4 CDM on redshift.

PR #285, opened 2022-03-03
updates from working session

clairblacketer:


katy-sadowski:
@clairblacketer are these still changes you'd like to add in? 

i noticed in the current concept-level threshold files most measurements don't have high/low plausible values anymore (including the ones being modified here). not sure if on purpose, or maybe a bug? (i saw @MaximMoinat also mentioned this in Teams, not sure if it was discussed yet?)

clairblacketer:
@katy-sadowski thanks for checking. I will close this PR as we decided to go a different way for measurement checks

PR #292, opened 2022-03-28
Fixes #33: special characters in metadata

### LINKED ISSUE #33: special characters in metadata
When pulling the CDM source metadata, special characters encoded in latin1 are brought into the JSON file. When calling viewDqDashboard, the jsonlite call to read the JSON fails due to these characters. 

Need to re-encode to UTF-8?

AnthonyMolinaro:
Implemented the fix we've been using for post processing to address this issue.  The following lines were added to the executeDqChecks() function:

```r
saveEncoding <- getOption("encoding")
options("encoding" = "UTF-8")
# code execution #
options("encoding" = saveEncoding)
```
Tested successfully on redshift.

### DIFF
--- a/R/execution.R
+++ b/R/execution.R
@@ -199,6 +199,10 @@ executeDqChecks <- function(connectionDetails,
   stopifnot(is.null(checkNames) | is.character(checkNames), is.null(tablesToExclude) | is.character(tablesToExclude))
   stopifnot(is.character(cdmVersion))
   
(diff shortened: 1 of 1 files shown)


clairblacketer (approved):

clairblacketer:
@AnthonyMolinaro I am not sure I can change this but can you target develop instead of main?

Issue #272, opened 2022-01-04
Issue with a gender check (gender = 5)


clairblacketer:
From the DQD:

For a CONCEPT_ID 2617204 (CERVICAL OR VAGINAL CANCER SCREENING, PELVIC AND CLINICAL BREAST EXAMINATION), the number and percent of records associated with patients with an implausible gender (**correct gender = 5**).

clairblacketer:
@AnthonyMolinaro this one is pretty straightforward, please look in the v5.2.2, v5.3.1, and v5.4 concept level files and check the concept id 2617204 and make sure the gender concept id is 8532

Issue #267, opened 2021-12-03
isStandardValidConcept should have a subcategory


clairblacketer:
This should have the Value subcategory

clairblacketer:
@AnthonyMolinaro this one is a simple fix in the control files as well. Please go through all the check_descriptions files and add the "Value" subcategory to the isStandardValidConcept check.

PR #295, opened 2022-03-29
Fixes issues #244, #267, and #272

### LINKED ISSUE #244: Birth_datetime used in plausibleTemporalAfter check
The birth_datetime field is used for the plausibleTemporalAfter check with observation period and visits. This should be fixed so that it coalesces between birth_datetime and june 1 and the birth year.

AnthonyMolinaro:
Updated control files to coalesce birth_datetime and yob-06-01.  The extra casting was necessary to avoid errors (on redshift) because birth_datetime is a timestamp.


AnthonyMolinaro:
Attaching file to see the SQL generated from new control files for this check.

[plausibleTemporalAfter.txt](https://github.com/OHDSI/DataQualityDashboard/files/8375027/plausibleTemporalAfter.txt)

clairblacketer:
closing this as they need to be done independently

PR #178, opened 2020-10-02
Field date between bounds

kinyusui:
Not sure if kahn context and categories were correct. Should I updated the v5.3.1 csv files let me know if I should do the same for v5.2.2 before merging.

A more abstract version would be the following

SELECT COUNT_BIG(*) AS violations
    FROM @cdmDatabaseSchema.@cdmTableName AS cdmTable
    INNER JOIN @cdmDatabaseSchema.@cdmTableName1 AS cdmTable1
    ON cdmTable.@connectorField = vo.@connectorField
    AND (
        cdmTable.@cdmFieldName < cdmTable1.@upperBoundDateField
        OR cdmTable.@cdmFieldName > cdmTable1.@lowerBoundDateField
    )

However that would require cdmTable1, connectorField, upperBoundDateField, and lowerBoundDateField as new parameters.

PR #300, opened 2022-03-30
New field level check: withinVisitDates

AnthonyMolinaro:
This check was originally introduced in [PR 178](https://github.com/OHDSI/DataQualityDashboard/pull/178) by @kinyusui.
The check determines if event dates occur within their corresponding visit dates.

The check has been renamed to withinVisitDates and the necessary control files were updated for 5.3.1 and 5.4.  This was successfully tested against 5.3.1 and 5.4 on redshift.


AnthonyMolinaro:
Moved to PR #306

PR #313, opened 2022-04-15
Domain and Class checks denominator only non-zero concepts

### LINKED ISSUE #309: fkDomain denominator rows over non-zero concepts.
I came across the following:
All mapped concepts in the visit occurrence table were from the wrong domain. However, the DQD reported only 40% of rows violated the check. It turned out the other 60% were rows with unmapped standared concept (mapped to 0). I would expect in this case that the fkDomain check would report 100% of concepts in the wrong domain.

Proposal:
Also filter denominator to non-zero conceps (`WHERE co.concept_id != 0`) here:
https://github.com/OHDSI/DataQualityDashboard/blob/edd8125ad8496fc9ba407cc7a44a473a57ac5767/inst/sql/sql_server/field_fk_domain.sql#L39-L47

There might be other checks where the same holds true.

MaximMoinat:
Fixes #309

The idea is that this change will give a more meaningful fail%. However, it might be confusing that the denominator count is different in the same table for different checks. Simplicity/consistency might be preferable. Happy to discuss.

- [ ] also apply for the `FIELD_IS_STANDARD_VALID_CONCEPT` check?

### DIFF
--- a/inst/sql/sql_server/field_fk_class.sql
+++ b/inst/sql/sql_server/field_fk_class.sql
@@ -43,5 +43,6 @@ FROM
     	ON cdmTable.PERSON_ID = c.SUBJECT_ID
     	AND c.COHORT_DEFINITION_ID = @cohortDefinitionId
     	}
+        WHERE cdmTable.@cdmFieldName != 0 
 ) denominator
 ;
--- a/inst/sql/sql_server/field_fk_domain.sql
+++ b/inst/sql/sql_server/field_fk_domain.sql
@@ -44,5 +44,6 @@ FROM
     	ON cdmTable.PERSON_ID = c.SUBJECT_ID
     	AND c.COHORT_DEFINITION_ID = @cohortDefinitionId
     	}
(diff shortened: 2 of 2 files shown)


clairblacketer:
@MaximMoinat I am not so concerned about differing denominators for the same table but I think we will need to be explicit about this check excluding concept_ids of zero. By definition, any zero concepts will fail the fk_domain check so we should discuss if that is something we want to remove. The zeros will be highlighted in some of the mapping completeness checks so I am fine with it if the group us. We can discuss at the June 2nd meeting.

MaximMoinat:
I see your point. I will not join the meeting as I am on holiday. Will join in two weeks to discuss!

clairblacketer:
> I see your point and happy to discuss. However, I not join the meeting as I am on holiday. But I can join in two weeks to discuss! @clairblacketer

Have fun on holiday!

MaximMoinat:
Closing to keep check as is. This will keep things simpler to explain: the %fail will represent the % of records in the table with a wrong domain, among all records in the table.

PR #346, opened 2022-08-09
Add "resume" option

ganisimov:
It happens that due to connection issues or other conditions executeDqChecks
fails and progress is lost. Resume option allows to recover progress of
previous execution.

When resume option is set to TRUE, and file with intermediate results of
individual check exists in outputFolder, processing of the check is
skipped and results are loaded from the file.

Once individual check is processed without resuming, intermediate results
are saved to file. When overall processing completes, additional files are
removed.


ganisimov:
Closing this PR due to need of more granular handling of saved check results.

Issue #109, opened 2020-03-23
Add ability to resume execution


alondhe:
For executions in which a check throws an error, it would be nice if we could re-run the check and integrate it into the results JSON file.

So, in executeDqChecks(), perhaps we add an optional field of resultsJsonFile, and with the checkNames parameter, can basically re-run the checks identified, and then we produce a new results JSON file that substitutes in the new results back into the original results JSON file.

clairblacketer:
+1 to this, we have some executions halted due to clusters being restarted and we lose all progress

PR #350, opened 2022-08-26
Add "resume" option

ganisimov:
It happens that due to connection issues or other conditions executeDqChecks fails and progress is lost. Resume option allows to recover progress of previous execution.

Once certain check is processed without error, it's saved to file corresponding to check description. If overall execution completes without errors these intermediate files are deleted.

When resume option is set to TRUE and intermediate results for certain check exist, processing of the check is skipped and saved results are used instead.


ganisimov:
This addresses #109

clairblacketer:
@ganisimov thank you for this! We recently refactored the R functions into separate files on the develop branch. Would you be able to move your changes to the executeDqChecks.R file on develop?

ganisimov:
@clairblacketer , thank you for looking into this! I created another PR #395.

PR #282, opened 2022-02-10
v2.0

dmitryilyn:
Includes all the changes from me and Luis.

PR #352, opened 2022-09-13
Add NA and Error status (v2)

MaximMoinat:
Simplifying the implementation of the check status from #282.

Two notable simplifications:
- This does not implement the 'isStructureRelatedError'. If the database throws a SQL error, it is flagged as an error and treated as not having passed.
- This does not force certain checks to be run (in v2 PR, the TABLE and FIELD level are forced to be run)

Items to implement:
- [ ] Add the cdmTable check, including the documentation on the 'Data Quality Check Type Definitions' page
- [ ] Update roxygen documentation (re-generate method reference)
- [ ] Add documentation on flags to the main documentation
- [ ] Warning if `cdmTable`, `cdmField` or `measureValueCompleteness` are not executed (this will make the NA check status less useful).


MaximMoinat:
To be implemented with #282

clairblacketer:
@MaximMoinat can I close this branch?

PR #358, opened 2022-09-15
V2 function refactor

clairblacketer:
This PR pulls out the individual functions from the execute.R file and puts them in separate files to be called.


clairblacketer:
FWIW, here is some testing I did. I first ran the code on the main branch on a local version of synthea I have. Here are the results:

image

3413 total checks (not withstanding the issue with the plausibility validation)

When I ran the exact same configuration using the code on develop I got:

image

3447 Total checks 

3447- 3413 = **34** extra checks

**21** are coming from the new cdmTable check that looks to see what tables actually exists to help calculate the checks that are not applicable

**8** are coming from the new withinVisitDates check

**5** are coming from the NOTE_ID column in the NOTE_NLP table. These are correct, they just did not have checks in previous versions.

MaximMoinat (approved):
I tested in my environment and results are the same as with the current develop branch

R/summarizeResults.R
```diff
@@ -0,0 +1,113 @@
...
+#' @param conceptChecks             A dataframe containing the concept checks
+#' @param metadata                  Information from the CDM_SOURCE table with details about the database
+#' 
+#' @keywords internal
+#' 
+
+.summarizeResults <- function(connectionDetails,
+                              cdmDatabaseSchema,
+                              checkResults,
+                              cdmSourceName,
```
MaximMoinat, line 22: Arguments `connectionDetails`, `cdmDatabaseSchema` and `cdmSourceName` are not used in this function.

R/runCheck.R
```diff
@@ -0,0 +1,83 @@
...
+                      connectionDetails,
+                      connection,
+                      cdmDatabaseSchema, 
+                      vocabDatabaseSchema,
+                      cohortDatabaseSchema,
+                      cohortDefinitionId,
+                      outputFolder, 
+                      sqlOnly) {
+  
+  library(magrittr)
```
MaximMoinat, line 31: `magrittr` is imported, no need to load here

MaximMoinat (approved):
See comments for a few suggestions for further cleanup

MaximMoinat:
Also, there is no need to use `DataQualityDashboard:::` to call the internal functions. These are simply available to the package.

It is even discouraged by CRAN: https://community.rstudio.com/t/how-to-include-internal-functions/38654/2

R/runCheck.R
MaximMoinat: My bad, upon testing, it seems like this is necessary.

R/summarizeResults.R
clairblacketer: Before I remove these in the PR we should test it. In one of the other functions it looked like some parameters were not used. When I removed them the whole thing broke. I think it has to do with the cluster apply and lapply usage throughout the package. They do some things under the covers that is not made explicit in the code.

PR #361, opened 2022-09-16
Display NA and Error on seperate lines in dashboard overview

MaximMoinat:
Making the dashboard overview compatible with both the old results and the v2 results with new check statuses.

For the v2 results, this PR correctly calculates the numbers in the overview, and adds three lines with: the number of NA checks, the number of Errors and a correct pass percentage.

image


clairblacketer:
Thanks @MaximMoinat! This is great!

PR #364, opened 2022-09-21
Add reEvaluateThresholds function

### LINKED ISSUE #290: Apply new thresholds without re-executing DQD
Currently, you have to run the whole DQD execution again after updating the thresholds to apply them. Like this:

```mermaid
graph LR
    A[Execute DQD on CDM] --> B[Inspect Results<br>in Shiny Viewer]
    B --> C[Edit Thresholds File]
    C --> A
```

If the CDM data did not change, you can actually determine which checks fail with the new thresholds, without re-executing DQD. This will make iterations faster. (yes, I am trying out [a new feature I recently learned about](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/))

```mermaid
graph LR
    A[Execute DQD on CDM] --> B[Inspect Results<br>in Shiny Viewer]
    B --> C[Edit Thresholds File]
    C --> B
```

I see two potential ways to implement this, the first requiring less work than the second:

- **Minor change**: Create a separate method that takes an DQD output json and new thresholds file and calculates which checks pass/fail given those new thresholds.
- **Major change**:  Redesign DQD execute to output only the raw data for each check (mainly `NUM_VIOLATED_ROWS` and `NUM_DENOMINATOR_ROWS`). Then, the viewer takes this raw output AND the thresholds to determine which checks have failed. Added benefit is that we can reduce the size of the results json.

MaximMoinat:
Fixes #290

The `reEvaluateThresholds` takes an old DQD json result and evaluates this against given thresholds, to creates a new DQD json result. This updates the pass/fail status. 

This PR also contains some additional refactorings on top of the v2_function_refactor branch.


clairblacketer:
@MaximMoinat you beat me to the punch with your bug fixes! I found the same things. I did also put a `toupper()` command on the concept check file just to be safe. I will send a commit with that change

PR #374, opened 2022-10-15
Styler changes

katy-sadowski:
Ran styler; this is what it did.


katy-sadowski:
Katy & Ajit inspected all changes, they seem OK!

MaximMoinat (approved):

PR #380, opened 2022-10-15
Enable Unit test for concept level

MaximMoinat:
Runs on subset of concept checks.


katy-sadowski:
i tested this locally and it works.  the check failures are coming from a weird folder that appeared in `sql-formatting` and is removed in #383 .

PR #389, opened 2022-10-16
Unit tests

schuemie:
Cleaned up the unit tests and added some more to improve coverage. Currently failing because of what seems like bugs in `reEvaluateThresholds()`


fdefalco (approved):

MaximMoinat:
@schuemie I cannot find the issue you described. Could you point to the code, then I can have a look.

PR #394, opened 2022-10-18
Unit Test Fixes

katy-sadowski:
- Added `expect_warning` wrapper to a few more `executeDqChecks` tests
- Consolidated test helper files


katy-sadowski:
Note: I used `expect_warning` rather than suppressing the warnings because that's what Martijn used in the unit test he added.  But I'm open to either approach if we think it's better to suppress vs running the expect warning test over and over.

PR #403, opened 2022-11-17
Add check to verify cdmVersion is valid

### LINKED ISSUE #399: Better error handling for bad argument of `cdmVersion`
When passing in a bad argument for `cdmVersion` in `executeDqChecks` (for example, an overly specific "5.4.0"), we have a cryptic (to the user) error about not being able to load a file, as shown below.

```
Error in read.table(file = file, header = header, sep = sep, quote = quote,  :
  no lines available in input
In addition: Warning message:
In file(file, "rt") :
  file("") only supports open = "w+" and open = "w+b": using the former
```

It looks like this comes from something like loading the threadhold CSVs here:

https://github.com/OHDSI/DataQualityDashboard/blob/main/R/executeDqChecks.R#L164

but I'm not positive.

While the allowed values are documented in the docstring for the function, there's no error handling that could help the user figure out what to do to fix this.

danielh2942:
this fixes #399 
also contains some small changes to make my linter quiet


R/executeDqChecks.R
```diff
@@ -13,8 +13,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-
+library(stringr)
```
azimov, line 16: Avoiding use of `library` or `require` is [required](https://ohdsi.github.io/Hades/developerGuidelines.html#Function_calls_must_not_have_invisible_side_effects) as this will override any definitions in the user space.
Use namespace `stringr::<function>` instead

R/executeDqChecks.R
```diff
@@ -81,6 +81,10 @@ executeDqChecks <- function(connectionDetails,
     stop("connectionDetails must be an object of class 'connectionDetails'.")
   }
 
+  if (! str_detect(cdmVersion, regex("^5.[2-4]$"))) {
```
azimov, line 84: adding the regular expression somewhere in the package would make this easier to modify e.g. `ACCEPTED_CDM_REGEX <- "^5.[2-4]$"`

azimov (changes requested):

R/executeDqChecks.R
danielh2942: Cool will do!

azimov (approved):

katy-sadowski:
Thanks @danielh2942 !  Merging this into `develop` branch; it'll ship with the upcoming release.

Issue #408, opened 2022-12-12
Error in OMOP_CDMv5.4_Field_Level.csv about offset field check


mbucalo:
As indicated in the title I found that in the OMOP_CDMv5.4_Field_Level.csv at line 316 the field offset is surrounded by multiple double quotes that leads the DQD to use it in the query with all capital letters. Removing the quotes allow the DQD to check correctly the NOTE_NLP table

katy-sadowski:
oooh this is fascinating.  i think the field name has quotes since `offset` is a reserved word in SQL.  but when you surround a column name in quotes (in Postgres at least), it makes the column name case sensitive.  however because DQD makes column names [all uppercase](https://github.com/OHDSI/DataQualityDashboard/blob/6c9c8fedb5a545dda6c34112acef40decda1fc5e/R/executeDqChecks.R#L234), it's looking for the "OFFSET" column, which doesn't exist.

it actually looks like this was [already reported](https://github.com/OHDSI/DataQualityDashboard/issues/245) and fixed, by just skipping this field from evaluation by DQD: https://github.com/OHDSI/DataQualityDashboard/blob/6c9c8fedb5a545dda6c34112acef40decda1fc5e/R/executeDqChecks.R#L206.  but the v5.4 thresholds file still has quotes around "offset" so the quoted "offset" isn't getting skipped by this line.  

i think the solution will either be to remove these quotes in the 5.4 thresholds file _or_ revisit the decision to make all column names uppercase (which was done [here](https://github.com/OHDSI/DataQualityDashboard/pull/101), seemingly in response to the same issue, but for Oracle, so not sure there's a universal fix possible...).  i will check with the team on this and get back to you with a fix!

katy-sadowski:
FYI - I chatted with the team and they've already tried to tackle this from multiple angles. The best solution until that column is renamed in CDM will be to skip it from evaluation in the DQD. I'll include the fix to remove the quotes from 5.4 threshold file in our upcoming release.

Thanks for reporting :)

mbucalo:
Thanks to you Katy for the update.

--
Mauro Bucalo
CTO & Product Manager | Biomeris S.r.l
T. +39 0382985055 | M. +39 3284362881

Il 20 dic 2022, 23:43 +0100, Katy Sadowski ***@***.***>, ha scritto:

katy-sadowski:
This is now resolved in the v2.1 release!

Issue #396, opened 2022-10-21
vignettes/AddNewCheck.rmd Add a New Check missing images


don-torok:
Images, Figure 1: launchSqlRenderDeveloper and Figure 2: launchSqlRenderDeveloper do not display in the rendered vignette.  The images are in vignettes/images and referenced as ![Figure 1: launchSqlRenderDeveloper](images/library_SQLRender.png)
 and ![Figure 2: launchSqlRenderDeveloper](images/SQLRenderDeveloper.png) in the markdown file AddNewCheck.md.  I did not write the .rmd file.

katy-sadowski:
weird issue! somehow this didn't get fixed after all. right now the website is pulling images from [here](https://github.com/OHDSI/DataQualityDashboard/tree/main/docs/articles/images), where the images have lowercase png but the path in the html file has uppercase PNG. so the images are broken on the website still.  and the .md files in the vignette folder still have lowercase png while the image files in that directory have uppercase PNG. haha.

i'll see if i can go through and make _everything_ lowercase so we can avoid this from causing issues moving forward (it's especially confusing because git diffs and local rendering don't seem to care about case of the suffix).

clairblacketer:
@katy-sadowski if this was fixed on the develop branch that might be why it is still broken. The website runs off of the docs folder on the main branch. It might be worth checking once you merge develop with main to release v2.0.

katy-sadowski:
i merged in the develop branch yesterday (v2 release is up!) and then verified this thinking i could close it. but unfortunately it's broken, probably because there's a mismatch in case between the html and the image in the docs/articles directory.

katy-sadowski:
This is resolved in the v2.1 release.

Issue #351, opened 2022-08-30
Add DataQualityDashboard to Hades


ablack3:
This is an issue for the 2022 Hades Hack-a-thon.

Hades requirements: https://ohdsi.github.io/Hades/packageRequirements.html

<detailed tasks to be added to this issue>

katy-sadowski:
https://docs.google.com/spreadsheets/d/1UnGJkLgpq52d6Q9LYiBAxUhCZBKkQiBiqKFisFZWi-c/edit?usp=sharing

anthonysena:
[HADES Hack-a-thon]: Katy is going to work on this item.

alondhe:
Most tasks are complete. Last item for the Sunday session: unit tests.

schuemie:
Is DQD ready for review for inclusion in HADES?

katy-sadowski:
PR #398 contains the final changes needed to meet the Hades requirements.  Once it's merged into main we should be good to go!  I can comment back here when that's the case.

schuemie:
Just checking in if you need any help getting this over the finish line.

katy-sadowski:
Thanks for checking in & apologies for the delay @schuemie.  Snakecase-->Camelcase change should be finalized and ready for review in the above PR this week - I'll tag you on that so you can follow along.  While that's being reviewed I'll work on the other items.

katy-sadowski:
Hi @schuemie !  As an FYI, here is the PR with the snakecase --> camelcase conversion: https://github.com/OHDSI/DataQualityDashboard/pull/409.  The other changes in the PR are some bug fixes I'm going to include in the next release along with the Hades requirements.  I will now work on the remaining Hades reqs and put up some more PRs for those in the coming week.  Thanks for your patience 🐫

katy-sadowski:
Hi again @schuemie 👋 I could use your advice on 2 of the R CMD Check notes.  The package is way over the max size (which I believe is 5 Mb), and some of the filenames are too long (ex. `storing paths of more than 100 bytes is not portable: ‘DataQualityDashboard/inst/shinyApps/www/vendor/googlefonts/sairaextracondensed_v11_-nFvOHYr-vcC7h8MklGBkrvmUG9rbpkisrTrU23h2wpk.ttf’`).  The main offender in both cases is the `vendor` folder for the shiny app: https://github.com/OHDSI/DataQualityDashboard/tree/main/inst/shinyApps/www/vendor

[...]

PR #409, opened 2022-12-14
v2.1 small fixes & add'l HADES requirements

katy-sadowski:
This PR contains the following changes which are planned to be released in [v2.1](https://github.com/OHDSI/DataQualityDashboard/projects/5) of DQD:

- **CDM Datatype Check:** Check verifies if integer fields contain only integers. However, right now it's actually checking if these fields are numeric, so pass rate could be inflated if there are float values in integer fields. I've continued the work started in #334 to fix this
  - Commits: eaa106eaf8bdbbf15b96005bc252b77860a12a26, a3bd9a4b02455115639e48cefa231f78829cd905
- **offset column:** Removed quotes from column name in `cdmFieldName` to ensure this column is skipped by DQD. See #408 
  - Commit: d1e0bf58f2905a5414fdee71429bbdf924b0d37a 
- **Vignette broken images:** Fixed case of png extensions so that images in `addNewCheck` render properly. See #396 
  - Commits: 02fc24af55bd5df626710ff04cbe0084af4c83e5, 1549a8eb3ea73ebf750591fea3fab4cd99e6b614
- **Snake case to camel case:** Converted all snake case variable/column names to camel case per HADES style guide requirement. See #351 
  - Commits: 98bcf3c8fbe56e510619965ca0824f801872b3e6, f7391be2004fc3816e12c4534a42c37f0560908c, 6264fd1411ef9652cc821c36c9637c3782c3f869, 357b8da0072868c7be77516a2f7509bca519847c, b331ee167f0b28614b558e19125bc0c3a59d574d

This PR ballooned a bit; I apologize for its size. I've listed the commits relevant to each bug/issue above to help with a smoother review. In the future I'll make separate PRs per issue! Also note a good chunk of the changed files are regenerated docs and/or deletion of extraneous docs files.


R/writeJsonResultsTo.R
```diff
@@ -73,6 +72,16 @@ writeJsonResultsToTable <- function(connectionDetails,
...
+  # convert column names to snake case, omitting the autogenerated rowname column
+  # and the checkId column, which has no underscore in the results table DDL
+  for (i in 1:ncol(df)){
+    if (colnames(df)[i] %in% c("X_row","checkId")) {
+      colnames(df)[i] <- tolower(colnames(df)[i])
+    } else {
+      colnames(df)[i] <- SqlRender::camelCaseToSnakeCase(colnames(df)[i])
+    }
+  }
+
```
katy-sadowski, line 84: A bit more color on this change.  Basically there is a column named `checkid` in the check results DDL queries ([example](https://github.com/OHDSI/DataQualityDashboard/blob/6c9c8fedb5a545dda6c34112acef40decda1fc5e/inst/sql/sql_server/result_table_ddl_field.sql#L21)) which should really be `check_id` to make for a consistent snakecase<>camelcase conversion.  However, I did not want to modify the DDL since I think that doing so could be considered a breaking change for some users who use the check results database table in downstream applications.

Regarding the `x_row` column - @clairblacketer do you know if this column serves any purpose in the code (searching the repo doesn't reveal any explicit references)?  I think it originates from a dataframe rowname column somewhere along the chain in the generation of the `checkResults` object (the X is pre-pended by R to make `_row` a valid colname).  If we apply camelcaseToSnakecase to this column, it will generate a different result than is in the DDL ("_x_row").  I was similarly hesitant to remove/change this col in the DDL in case that would be a breaking change for any users.

Both of the above mean we need this sorta clunky workaround to enable the camel to snakecase conversion.  I will definitely file an issue to clean this up in the next major release.

What do you think? 

Note there's a similar change on line 76 of `writeResultsTo.R`

katy-sadowski:
@clairblacketer FYI I just snuck in one more commit here (d16399c1b4c3d91d42abd7acb0a55a7eb9909371).  It's to resolve the R CMD Check note `no visible binding for global variable` which we were getting for several dataframe column references and which Martijn describes in more detail [here](https://github.com/OHDSI/DataQualityDashboard/issues/351#issuecomment-1346123314).  It was cleaner to include this change alongside the snakecase-->camelcase change because all of the variables in question are converted to camelcase in this PR.

Additionally, the commit includes a change to resolve a dependency issue which just popped up due to the new DatabaseConnector release. I've put a comment on the code to flag that as I need input as to whether this is the right approach to address the issue.

Thanks very much for your review!  I'd be happy to pair on this on a quick call if it'd be helpful since this PR is quite hefty.  Let me know :)

R/executeDqChecks.R
```diff
@@ -77,11 +78,11 @@ executeDqChecks <- function(connectionDetails,
                             fieldCheckThresholdLoc = "default",
                             conceptCheckThresholdLoc = "default") {
   # Check input -------------------------------------------------------------------------------------------------------------------
-  if (!("connectionDetails" %in% class(connectionDetails))) {
-    stop("connectionDetails must be an object of class 'connectionDetails'.")
+  if (!any(class(connectionDetails) %in% c("connectionDetails", "ConnectionDetails"))) {
+    stop("connectionDetails must be an object of class 'connectionDetails' or 'ConnectionDetails'.")
```
katy-sadowski, line 82: @schuemie the recent DatabaseConnector upgrade was causing `executeDqChecks` to fail at this input check, as it seems the `connectionDetails` created by the new DatabaseConnector version is of `ConnectionDetails` class, not `connectionDetails`?  

my change here was inspired by your Eunomia PR [here](https://github.com/OHDSI/Eunomia/pull/36).  is this OK?  or should we require DatabaseConnector >6.0.0?  or something else?  i'm not familiar with what the change in capitalization represents so want to make sure i do the right thing here - thanks.

schuemie: The change you have is fine. Sorry about that. I had no idea that simple renaming would have such wide impact.

clairblacketer (approved):

PR #411, opened 2022-12-19
Add ability to resume execution of DQD checks

ganisimov:
It happens that due to connection issues or other conditions executeDqChecks fails and progress is lost.

When new resume argument is set to TRUE, file in outputFile path is used as source of check results instead of actual processing. Missing check results or results reporting error are re-processed.

Note that outputFile must be set explicitely to make resuming to take effect. New results overwrite file in outputFile path.

Addresses #109

Before you do a pull request, you should always **file an issue** and make sure the package maintainer agrees that it's a problem, and is happy with your basic proposal for fixing it. We don't want you to spend a bunch of time on something that we don't think is a good idea.

Additional requirements for pull requests:

-   Adhere to the [Developer Guidelines](https://ohdsi.github.io/Hades/developerGuidelines.html) as well as the [OHDSI Code Style](https://ohdsi.github.io/Hades/codeStyle.html).

-   If possible, add unit tests for new functionality you add.

-   Restrict your pull request to solving the issue at hand. Do not try to 'improve' parts of the code that are not related to the issue. If you feel other parts of the code need better organization, create a separate issue for that.

-   Make sure you pass R check without errors and warnings before submitting.

-   Always target the `develop` branch, and make sure you are up-to-date with the develop branch.

### DIFF
--- a/R/runCheck.R
+++ b/R/runCheck.R
@@ -27,7 +27,9 @@
 #' @param cohortDatabaseSchema      The schema where the cohort table is located.
 #' @param cohortDefinitionId        The cohort definition id for the cohort you wish to run the DQD on. The package assumes a standard OHDSI cohort table called 'Cohort'
 #' @param outputFolder              The folder to output logs and SQL files to
+#' @param outputFile                (OPTIONAL) File to re-use results of previous execution if resume is set
 #' @param sqlOnly                   Should the SQLs be executed (FALSE) or just returned (TRUE)?
+#' @param resume                    Boolean to indicate if processing will be resumed
 #'
 #' @import magrittr
 #'
@@ -44,7 +46,9 @@
                       cohortDatabaseSchema,
                       cohortDefinitionId,
                       outputFolder,
-                      sqlOnly) {
+                      outputFile = "",
+                      sqlOnly,
+                      resume) {
   ParallelLogger::logInfo(sprintf("Processing check description: %s", checkDescription$checkName))
 
   filterExpression <- sprintf(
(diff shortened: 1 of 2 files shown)


R/runCheck.R
```diff
@@ -92,16 +111,52 @@
...
+        if (!is.null(checkResultsSaved)) {
+          currentCheckId <- .getCheckId(
+            checkLevel = checkDescription$checkLevel,
+            checkName = checkDescription$checkName,
+            cdmTableName = check["cdmTableName"],
+            cdmFieldName = check["cdmFieldName"],
+            conceptId = check["conceptId"],
+            unitConceptId = check["unitConceptId"]
+          )
+          checkResultCandidates <- checkResultsSaved %>% dplyr::filter(checkId == currentCheckId & is.na(ERROR))
```
MaximMoinat, line 124: This assumes the `ERROR` field is present in the results object, which is not the case when none of the checks resulted in an error.

MaximMoinat (changes requested):
This is a great addition, thanks. See my comment in-line.

Two remarks:
 - To be consistent with other OHDSI packages, I propose to either call this `incremental` mode like in [CohortDiagnostics](https://ohdsi.github.io/CohortDiagnostics/reference/executeDiagnostics.html) or add a new function `runMissingChecks` like in [Achilles](https://ohdsi.github.io/Achilles/reference/runMissingAnalyses.html). @katy-sadowski Maybe something to discuss in an upcoming WG meet?
 - Does this also work if a fatal error occurs? (e.g. connection lost, or R session timing out). I think the json results object only gets written at the end, so we might need to write out intermediate results to make this work for all cases where a DQD execution might be interrupted.

katy-sadowski:
I 100% want to add this feature, thanks @ganisimov for the contribution and @MaximMoinat for reviving the thread 😃 

I've been thinking, however, that we might want to wait to add it in as part of the larger check-running workflow overhaul we've been discussing implementing in 2024.  Part of the vision for that workflow includes moving away from the current "all-or-nothing" execution towards something more incremental that's aware of the relationships/dependencies between checks.  I think that implementation will lend itself better to figuring out how/where to store incremental results, how to handle various failure modes, etc.

Would definitely love to discuss this in a WG meeting early next year!

PR #417, opened 2023-02-13
v2.1 release

katy-sadowski:
Contains changes from #409 as well as doc updates, etc. required for release (commit 607d2e98818c7fd6bb5de6aa3340db44c8d282dc).


katy-sadowski:
@clairblacketer just an FYI on the recent action here - connection to the remote Oracle database was hanging, causing our automated checks to fail.  I commented out Oracle from the list of remote DBs to run tests against and now all checks are passing as you can see 🙏 I'll ask about the Oracle DB to see what's up, but I don't think this should block merge as it seems to be a connection issue rather than something with the DQD code.

katy-sadowski:
Merged in filename fixes from #414 to round out all remaining HADES requirements in this release!

clairblacketer (approved):

PR #440, opened 2023-04-26
Add function to convert json results params between snakecase and camelcase

katy-sadowski:
### Background
In the v2.1.0 [release](https://github.com/OHDSI/DataQualityDashboard/blob/main/NEWS.md#dataqualitydashboard-210), we converted all variables from snakecase to camelcase in order to comply with HADES requirements. This included variables in the output results json files as well as the corresponding code for consuming results files in the Shiny app.

As a result, new versions of DQD cannot read pre-v2.1.0 DQD files.  Additionally, the Ares package expects snakecase files and thus cannot currently support results files produced by newer DQD versions.

### Changes in this PR

- Adds a public function which can be used to convert results json files between snake and camelcase. Users and other applications can use the function to convert their files as needed. Additionally, the DQD Shiny app can use the function to convert the case of older results files so that users don't need to worry about the case of their files when launching the viewer
- Updates app.R to convert snakecase files before reading in results and launching app


R/convertResultsCase.R
```diff
@@ -0,0 +1,79 @@
...
+  if (targetCase == "camel") {
+    swapFunction <- SqlRender::snakeCaseToCamelCase
+  } else {
+    swapFunction <- function(x) {
+      toupper(SqlRender::camelCaseToSnakeCase(x))
+    }
+  }
+  
+  results$Metadata <- dplyr::rename_with(results$Metadata, swapFunction)
+  results$CheckResults <- dplyr::rename_with(results$CheckResults, swapFunction, -c("checkId"))
```
katy-sadowski, line 68: the checkId variable was already camelcase in the older DQD versions

fdefalco (approved):
I was able to verify this is functioning as intended in my test environment to convert new camel case to snake case to allow functioning with existing AresIndexer and Ares packages.

katy-sadowski:
Thanks for the review, @fdefalco !  I'll ship this with the next release!

Issue #436, opened 2023-04-11
Error for 2 plausible unit concept checks


katy-sadowski:
```The number and percent of records for a given CONCEPT_ID 3036277 (BODY HEIGHT) with implausible units (i.e., UNIT_CONCEPT_ID NOT IN ( INF)). (Threshold=5%).```

```The number and percent of records for a given CONCEPT_ID 3023540 (BODY HEIGHT MEASURED) with implausible units (i.e., UNIT_CONCEPT_ID NOT IN ( INF)). (Threshold=5%).```

For some reason the list of unit concept IDs is not getting parsed properly for these 2 checks.

MaximMoinat:
This could be due to `readr` v2 handling type guessing change: https://readr.tidyverse.org/articles/column-types.html#legacy-behavior 
https://readr.tidyverse.org/articles/column-types.html#legacy-behavior

Issue #367, opened 2022-09-30
Function to generate result json from results in database


edward-burn:
At the moment I don't think there is any easy way to generate the result json from the results saved in the database. I was in this situation with someone else having run DQD and so I can see the results in the database but could not see a way to get the json results without having to rerun the analysis. That was until @MaximMoinat shared some bespoke code that meant I could do this. It would I think be helpful to make this a function of the package if possible.

katy-sadowski:
Hey @edward-burn - that code you mention has now been added in a vignette [here](https://ohdsi.github.io/DataQualityDashboard/articles/SqlOnly.html#optional-execute-queries) (see code starting at `# Get results`)

jshoughtaling:
Hey @edward-burn and @katy-sadowski - I opened PR #497 yesterday that addresses this issue and adds a public function + tests in line with the vignette linked above.

katy-sadowski:
This is added in v2.5! Thanks @jshoughtaling !

Issue #333, opened 2022-06-10
Efficient SQL-only option to insert all results into dqdashboard_results table


TMSWhite:
As-is, the sqlOnly option has 4 main limitations:

- It requires connectivity to a live OMOP instance
- It does not include the metadata about the test being run
- It does not insert the results into the dqdashboard_results table
- It is inefficient, running all ~3000 tests serially rather than in parallel

[Pull request # 301](https://github.com/OHDSI/DataQualityDashboard/pull/301) solves all of these issues.  Users can specify the # of tests to run in parallel, and they are composed into a CTE before the full set of them are inserted into the output table.  In tests on Spark with 4 years of data and 3 million patients, performance improved from 16 hours to run all tests serially to about an hour to run them on a Spark cluster (e.g. DataBricks or HDInsight) and insert all results into dqdashboard_results.

katy-sadowski:
This has officially been released!  Thanks so much @TomWhite-MedStar for contributing this code!!!  I enjoyed the collaboration and am excited to see how this new addition will benefit our community.

It'd be great to have you attend a Data Quality workgroup meeting to present the feature - we are actually meeting this Thursday 5/25 at 9:30am eastern time in case that works for you.  Otherwise we can choose another one in the future.  Several members already expressed interest in testing out the feature; I'm looking forward to getting their input & performance benchmarks :)

PR #443, opened 2023-04-29
Fix sql only; add sqlOnlyUnionCount 

TomWhite-MedStar:
(1) When using sqlOnly, does not require actual database connection
(2) Generated SQL inserts each query result into dqdashboard_results table with all associated metadata; and computes threshold violations as part of each query (rather than waiting until all query results have been run).
(3) sqlOnlyUnionCount option lets multiple DQD checks be run in parallel (as a large cte).  This can reduced significantly reduce processing time.

Note, presumes that user has pre-created the dqdashboard_results table prior to running the SQL.
The output of this has been tested and validated on Databricks, and HDInsight-flavors of Spark SQL.


TomWhite-MedStar:
@MaximMoinat  and @clairblacketer , this is a replacement pull request for #301 .  I had to switch development forks, which I hope doesn't introduce any confusion.

TomWhite-MedStar:
OK, it isn't quite ready yet.  Although the queries run, there are two inconsistencies:
1. plausibleUnitConceptIds - these aren't being passed correctly to .createSqlOnlyQueries()
2. Plausible Value Low (and High) are generated at the Field level, but not at the concept+unit level

TomWhite-MedStar:
OK - this is working now.
The issue I was seeing with plausibleUnitConceptIds  was actually issue #436, which I fixed separately with pull request #444

I'll propose merging back in the missing plausibleValueHigh and plausibleValueLow tests via a separate pull request.

MaximMoinat:
Thanks @TomWhite-MedStar for making these fixes. It works for me, tested on v5.4 and postgresql.

Final step would be to make a vignette describing the functionality, with its benefits and limitations. But I would be happy to merge this already.

@katy-sadowski Would you like to do a final review and merge? And do we require to write a unit test for this?

TomWhite-MedStar:
@MaximMoinat , I updated the Getting Started vignette to show how to use sqlOnlyUnionCount, along with benefits and limitations.

katy-sadowski:
> @katy-sadowski Would you like to do a final review and merge? And do we require to write a unit test for this?

Yes, I'd like to do a review here to make sure I have a good grasp on the change & decide how it should fit into the release cycle.  I'll reply back here with my review.  Thanks so much for working on this @TomWhite-MedStar & @MaximMoinat !!! 

In the meantime, yes, please do add unit test(s) to [test-execute](https://github.com/OHDSI/DataQualityDashboard/blob/main/tests/testthat/test-execute.R) for the new functionality.  We don't yet have strong conventions for unit testing (that's my next big project 😉); if you just roughly follow the approach taken by other tests in there that's fine.

TomWhite-MedStar:
@katy-sadowski , I fixed the issue with the failing tests, and added a new test for sqlOnly and sqlOnlyUnionCount.

Good luck as you add more unit tests.  Now that I understand it, it is pretty convenient to use - but I'm sure you'll want more test coverage.

TomWhite-MedStar:
By the way, this pull request also solves issue #333

TomWhite-MedStar:
As benchmark, using sqlOnlyUnionCount = 100, I just ran all 3346 tests in 21 minutes on a small Databricks spark cluster.  Our measurement and observation tables each have about 5B records.  By comparison, when I last ran DQD without the sqlOnlyUnionCount option about a year ago on less than half the data volume, it took 16 hours on a similar Databricks cluster.  So, at least in our environment, having this option yields a 50-100 fold speed improvement.

inst/sql/sql_server/result_dataframe_ddl.sql
```diff
@@ -1,9 +1,9 @@
 --DDL to create dqdashboard_results table.
 
-IF OBJECT_ID('@resultsDatabaseSchema.@tableName', 'U') IS NOT NULL
-	DROP TABLE @resultsDatabaseSchema.@tableName;
+IF OBJECT_ID('@tableName', 'U') IS NOT NULL
+	DROP TABLE @tableName;
```
MaximMoinat: Why would we drop the schema placeholder? I would like to be able to specify where the results tabel is created.

inst/sql/sql_server/sqlOnly/cte_sql_for_results_table.sql
```diff
@@ -0,0 +1,35 @@
...
+  ,'' as warning
+  ,'' as error
+  ,'@checkid' as checkid
+  ,0 as is_error
+  ,0 as not_applicable
+  ,CASE WHEN (cte.pct_violated_rows * 100) > @threshold_value THEN 1 ELSE 0 END as failed
+  ,CASE WHEN (cte.pct_violated_rows * 100) <= @threshold_value THEN 1 ELSE 0 END as passed
+  ,NULL as not_applicable_reason
+  ,@threshold_value as threshold_value
+  ,'' as notes_value
```
MaximMoinat, line 32:
```suggestion
  ,NULL as notes_value
```
An empty string will be shown in dashboard as if note exists.

inst/sql/sql_server/result_dataframe_ddl.sql
TomWhite-MedStar: This is to  fix failing unit tests.
In writeResultsTo.R, tableName is defined in lines 55-58 as concatenation of @resultDataBaseSchema, @writeTableName, and @cohortDefinitionID (if needed). So, followed similar approach in sqlOnly.R line 134 (but without cohort option, since it doesn't yet use that).

MaximMoinat: I see, that should work then, thanks.

MaximMoinat:
Did some final testing, and results using SqlOnly-mode and regular DQD run are very similar. Same number of checks run, and (almost) same failed checks. Only differences was in five checks that have status NA in regular run, and fail in SqlOnly mode. This is expected.

@TomWhite-MedStar I created a separate article/vignette on the use of SqlOnly-mode with lots of footnotes and some example code. Could you merge this into your branch? https://github.com/TomWhite-MedStar/DataQualityDashboard/pull/3

TomWhite-MedStar:
@MaximMoinat , thanks for adding that vignette.  I made the requested changes.  

Is there anything else you need me to do?

R/executeDqChecks.R
```diff
@@ -127,7 +132,10 @@ executeDqChecks <- function(connectionDetails,
     metadata$dqdVersion <- as.character(packageVersion("DataQualityDashboard"))
     DatabaseConnector::disconnect(connection)
   } else {
-    metadata <- NA
+    metadata <- data.frame(
+      DQD_VERSION = as.character(packageVersion("DataQualityDashboard")),
```
katy-sadowski, line 136: here and throughout the PR, please convert all variable names to camelcase; this is a required convention for HADES packages

R/sqlOnly.R
```diff
@@ -0,0 +1,221 @@
+#' @file sqlOnly.R
```
katy-sadowski, line 1: please add a copyright header to this file (can copy from another .R file)

R/sqlOnly.R
```diff
@@ -0,0 +1,221 @@
...
+
+#' Internal function to create queries when running in SqlOnly mode
+#' 
+#' @noRd
```
katy-sadowski, line 5: please add `@param`s to these functions 🙏

R/sqlOnly.R
```diff
@@ -0,0 +1,221 @@
...
+    )
+  )
+}
+
+
+#' Internal function to evaluate one threshold
+#' Note: this does not evaluate is_error or not_applicable status
+#' @noRd
+#' @keywords internal
+.evaluateOneThresholdSqlOnly <- function(
```
katy-sadowski, line 157: should this be named "get threshold" instead, since it's not actually evaluating the threshold?

R/executeDqChecks.R
```diff
@@ -81,7 +83,7 @@ executeDqChecks <- function(connectionDetails,
                             fieldCheckThresholdLoc = "default",
                             conceptCheckThresholdLoc = "default") {
   # Check input -------------------------------------------------------------------------------------------------------------------
-  if (!any(class(connectionDetails) %in% c("connectionDetails", "ConnectionDetails"))) {
+  if (!sqlOnly && !any(class(connectionDetails) %in% c("connectionDetails", "ConnectionDetails"))) {
```
katy-sadowski, line 86: the connectiondetails object should still be valid in `sqlOnly` mode, right, even though we don't use it connect to the DB? since we use it to determine the dbms?


... that said I don't love the idea of using a connectiondetails just to get the dbms for sqlOnly mode. maybe we can add a dbms parameter that's required if sqlOnly is true, and make connectionDetails optional for sqlOnly?

R/sqlOnly.R
```diff
@@ -0,0 +1,221 @@
...
+    tableChecks,
+    fieldChecks,
+    conceptChecks,
+    sql,
+    connectionDetails,
+    checkDescription
+) {
+  qnum <<- qnum + 1
+  
+  check_description = SqlRender::render(
```
katy-sadowski, line 20: to simplify this a bit, could you actually just pass in `params` here like they do in `runCheck`? (i didn't try this so not sure it would work but seems it might?)

inst/sql/sql_server/sqlOnly/cte_sql_for_results_table.sql
```diff
@@ -0,0 +1,35 @@
...
+  ,'@unit_concept_id' as unit_concept_id
+  ,'@sql_file' as sql_file
+  ,'@category' as category
+  ,'@subcategory' as subcategory
+  ,'@context' as context
+  ,'' as warning
+  ,'' as error
+  ,'@checkid' as checkid
+  ,0 as is_error
+  ,0 as not_applicable
```
katy-sadowski, line 27: i don't _love_ that this approach doesn't use not_applicable status (but understand why). this makes its output inconsistent with what we'd get from DQD. what do you think about adding another function that the user can run to recalculate the not_applicable status?

i guess in this SQL only world that might be another SQL query that updates rows in the results table?

katy-sadowski (changes requested):
Hi @TomWhite-MedStar !  As I've started familiarizing myself with this code and the problems we're trying to solve with the new sqlOnly functionality, I have some thoughts I'd like to run by you & @MaximMoinat before we proceed. I also left some comments on the code with requested changes :) 

It seems that we're trying to achieve 2 separate goals with this PR:

- Improve DQD execution time
- Ability to use SQL output by sqlOnly to evaluate check results and write them to the database

I think each is probably worth its own serious consideration to come up with optimal solution(s) that are generalizable across as many users as possible.  I'm curious, for your use case @TomWhite-MedStar, are you seeing this sqlOnly path just as a workaround to DQD's slowness on your DB?  Or do you also have the need to separate out the R portion from the SQL/database connection?

That said...properly addressing DQD performance will be a complex and lengthy process given all the variables that can affect SQL query execution time and I don't want to block your contribution on that, especially if it could be useful for other users in the short term.  I also think it'd be interesting to have folks on other database systems help us generate some performance benchmarks for classic DQD vs the union'ed query approach.

So I'm open to adding this feature in an upcoming release!  I think it might be best to indicate it as a "beta" feature since we are not yet sure if/how this will be beneficial across the user base, and because putting it to use is rather complex.  I can ponder the best way to do this (probably just making the beta status more clear in docs, and building upon the vignette @MaximMoinat added) and update this branch accordingly, if that sounds good to you!

One other important thing we'll need to do in order to release this - as implemented, this is a breaking change as it removes the current sqlOnly functionality.  Can you please update the PR to include a parameter that specifies the sqlOnly "method", and allows for the previous sqlOnly approach to be used? (This should also be the default sqlOnly behavior. Basically users should be able to upgrade DQD and keep using sqlOnly the "old way" without changing anything about their setup.)

If it's easier to discuss all of this on a call, I'd be happy to set something up.  Thanks very much for your work on this; it's gotten me thinking a lot about performance and I think it'll be a good kickstart for future improvements to the package!

TomWhite-MedStar:
@katy-sadowski , thanks for the lengthy review & discussion.  I'll answer your code-specific questions separately.

You are mostly correct about the goals, but let me rephrase:
1. Fix bug in sqlOnly that requires live database access in order to generate SQL.  I didn't (and others may not) have access to live databases.  That requirement also prevents seeing the generated SQL for other database types (which is sometimes desirable).
2. Fix bug in sqlOnly that prevents the results of the queries from being inserted into the database -- by default, it just generates a long list of queries which by themselves are not useful.
3. Improve DQD execution time

Furthermore, the main reason I need this is that I will never be using R directly to run DQD (or Achilles).  Instead, I pre-generate the needed SQL and use DataBricks pipelines to run our full ETL (including Achilles and DQD).  That way we can manage the full ETL pipeline and ETL in our internal Git environment and more robustly support Dev/Test/Prod migrations.  So, I do need to `separate out the R portion from the SQL/database connection`

`So I'm open to adding this feature in an upcoming release!`.  Thanks!

When you say, `as implemented, this is a breaking change as it removes the current sqlOnly functionality`, I don't think that is accurate.  The first two goals above are actually bugfixes.  Nothing should break for existing users.  However, I'm fine with keeping it Beta and a separate CodeToRun example just in case.  I just can't imagine that anyone actually uses the generated sqlOnly output as-is.  I tried before submitting the PR, and it was totally non-viable for me.

@MaximMoinat went a level further than I did to show how to take the generated files from sqlOnly could be called using R after the fact.

I do think there is value down the road of switching the main code to use the new workflow in sqlOnly.R so that all SQL gets fully generated before being called.  That way, if people want to use R to use the Union feature without first generating the SQL, they can.  This would also help debugging, as it is currently quite hard to see the actual queries before being called.  Furthermore, some people would like to insert intermediate results as DQD is running, whereas others may want to wait and collect all results first and then insert them (as via the current  R workflow).  Most importantly, however, by merging those workflows, it avoid the risk of changing in the main R getting out of sync with the sqlOnly code.

All that being said, however, I'd advocate for including this PR as-is and worry about broader re-factoring at a later stage.

Thanks for all you do with managing this repo!

I'd be happy to get on a call to discuss further.

R/sqlOnly.R
TomWhite-MedStar: good idea - will do

tests/testthat/test-execute.R
```diff
@@ -272,3 +272,27 @@ test_that("Execute reEvaluateThresholds on Synthea/Eunomia", {
...
+      resultsDatabaseSchema = "@resultsDatabaseSchema",
+      cdmSourceName = "Eunomia",
+      checkNames = "measureValueCompleteness",
+      outputFolder = outputFolder,
+      writeToTable = TRUE,
+      sqlOnly = TRUE,
+      sqlOnlyUnionCount = 100,
+      writeTableName = "dqd_results"
+    ),
+    regexp = "^(Unknown or uninitialised column|Missing check names).*"
```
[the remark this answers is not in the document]
TomWhite-MedStar: will do

R/executeDqChecks.R
TomWhite-MedStar: I'll take a look.  Yes, in the sample, dbms is part of connectionDetails even if it is fake.  I agree with your concern - I'll see if it is easy to go that route (making dbms required and connectionDetail optional if using sqlOnly).

inst/sql/sql_server/sqlOnly/cte_sql_for_results_table.sql
TomWhite-MedStar: @katy-sadowski , not_applicable was implemented after I build the initial PR last last March.  Where should I look to understand how it is supposed to work?  Is it a matter of simply filtering out queries/checks that are NA?

R/executeDqChecks.R
TomWhite-MedStar: Which variables do you want as camelCase?  `metadata` is a pre-existing variable.

inst/sql/sql_server/sqlOnly/cte_sql_for_results_table.sql
TomWhite-MedStar: @katy-sadowski , is it fair to say that whenever num_denominator_rows = 0 then not_applicable should = 1?  If so, that is an easy change.  For sqlOnly, that should address all cases where there are missing concept or concept+unit combinations.  It would not address situations where a table or field are missing (for in those cases,  the query itself would fail).

R/executeDqChecks.R
TomWhite-MedStar: adding @dbms would require a brand new parameter.  I stuck with connectionDetails  but reverted line 84.

R/sqlOnly.R
TomWhite-MedStar: If OK, I'd rather keep this verbose so it is clear what parameters are passed.

TomWhite-MedStar:
@katy-sadowski , I made the changes you requested, other than a couple where I have additional questions.

R/executeDqChecks.R
TomWhite-MedStar: I think I figured out what you want.  Done

inst/sql/sql_server/sqlOnly/cte_sql_for_results_table.sql
TomWhite-MedStar: I modified this assuming that if num_denominator_rows = 0 then not_applicable = 1.  If that doesn't behave as expected, I can revert this.

TomWhite-MedStar:
I  completed all of the requested changes.

inst/sql/sql_server/sqlOnly/cte_sql_for_results_table.sql
MaximMoinat: While this is a good proxy, the regular NA status is a bit more complex and uses the result of other checks.
To avoid confusion, we should either exactly replicate the NA status or leave it out completely.

MaximMoinat:
Thanks for your reflections @katy-sadowski and @TomWhite-MedStar!

> When you say, as implemented, this is a breaking change as it removes the current sqlOnly functionality, I don't think that is accurate. The first two goals above are actually bugfixes. Nothing should break for existing users. However, I'm fine with keeping it Beta and a separate CodeToRun example just in case. I just can't imagine that anyone actually uses the generated sqlOnly output as-is. I tried before submitting the PR, and it was totally non-viable for me.

Agreeing with Tom here. Although any trivial change might break someone's pipeline, I think that in this case the chance is very small as described by Tom.

Making this a 'beta' feature makes a lot of sense to me. As indicated in the vignette I added, the output will look a bit different. So for most users I would NOT recommend using this sqlOnly feature.

> Furthermore, some people would like to insert intermediate results as DQD is running

That is another benefit I hadn't thought about. Having DQD run in 'incremental' mode with the ability to pick up after an error would potentially save a lot of time for people with huge databases and less powerful infrastructure.

> is it fair to say that whenever num_denominator_rows = 0 then not_applicable should = 1? If so, that is an easy change. For sqlOnly, that should address all cases where there are missing concept or concept+unit combinations.

While this is a good proxy, the regular NA status is a bit more complex and uses the result of other checks.
To avoid confusion, we should either exactly replicate the NA status or leave it out completely. I think this do-able with some post-processing in R. As indicated by Tom, let's leave that as a to-do for a later update. I can create an issue for this with some hints on how this could be solved.

TomWhite-MedStar:
@MaximMoinat , thanks.  I reverted to using 0 for NA.

A few other comments on the documentation.
1. I tried to follow the params syntax in the .sql queries - such as `queryText = @queryText`.  However, that broke the output since SQLRender does the substitution within the comments too -- so when queryText had embedded comments, substitution within `/**** ... @queryText ... ***/` would cause the comment blocks to prematurely terminate and the final SQL would be wrong.
2. I tried to insert the rendered queryText into the final output, but ran into a similar problem, so kept it as Query NN
3. I prototyped pre-computing the not_applicable_reason text (since it should be unique per check) so that it can be passed into the SQL, but given Maxim's comments that the logic is more complex and should be done after the fact, left it out.

Thanks again for all your help with this.

inst/sql/sql_server/sqlOnly/cte_sql_for_results_table.sql
TomWhite-MedStar: OK - reverted to 0 for NA

R/executeDqChecks.R
katy-sadowski: Thanks!  These 2 variables in the metadata should also be camelcase :)

R/sqlOnly.R
```diff
@@ -0,0 +1,296 @@
...
+    file = file.path(
+      outputFolder, 
+      "ddlDqdResults.sql"
+    )
+  )
+}
+
+
+#' Internal function to evaluate one threshold
+#' Note: this does not evaluate is_error or not_applicable status
```
katy-sadowski, line 218: update evaluate --> get

katy-sadowski (changes requested):
Suggested a couple more minor changes.  Also @TomWhite-MedStar please rebase your branch off of the latest version of `develop`; I merged some changes in yesterday which you'll want to incorporate here (minor ones!).

inst/sql/sql_server/sqlOnly/cte_sql_for_results_table.sql
katy-sadowski: 👍  yeah I was thinking an additional R function (or SQL query) that looks at the whole results table and then evaluates NA status. on board with @MaximMoinat 's suggestion below to tackle that in a future issue/PR.

R/executeDqChecks.R
katy-sadowski: OK yeah I think it's fine to leave as-is for now, either way is somewhat non-ideal so think it's best to keep it simpler.

katy-sadowski:
Thanks both for the additional discussion and @TomWhite-MedStar for the code changes!

Unfortunately I'm going to insist that we provide the (default) option to generate the old style SQL output files. The change is big enough that if anyone was using these files, it will break whatever workflow/code they have in place. In order to comply with semantic versioning conventions required by HADES, if we want to put this in a minor release we can't include any breaking changes. I'm certainly open to overhauling this feature in the next major release but we'll want to battle test this new approach and give users plenty of notice before we get there. @TomWhite-MedStar I'm picturing a `sqlOnlyMethod` parameter in `executeDqChecks` which is used to determine new vs old method - are you able to make that change? 

A couple more thoughts on the high level goals:

- Just to make sure I'm not missing anything here, I think the bug that required DB connection in sqlOnly mode has been fixed already, right? (Thinking of the [update](https://github.com/OHDSI/DataQualityDashboard/commit/2bec1029cfb5a5f77836be8f041e14d83a06ddd0) to skip querying CDM_SOURCE in sqlOnly mode)
- The extra context on how you integrate the DQD SQL into your ETL is very helpful @TomWhite-MedStar. It's great that this update will enable this for other users; I think it's a solid use case and consistent with other OHDSI tools where we provide multiple ways to run analyses. Curious if you use the DQD shiny app at all, or do you report on / utilize check results elsewhere? (Made me think of [this issue](https://github.com/OHDSI/DataQualityDashboard/issues/367) where a user wants to go from results table --> json --> Shiny app 😃 )
- Concur with both of your thoughts on future improvements this approach could enable in the main DQD functionality, especially around incremental insertion of results. We have another open issue around ability to resume DQD execution after an error which this could address. Another thought I had regarding improving performance was adding a multi-threaded execution option _within_ runCheck (essentially what we're simulating here with the UNIONs, I think). Let's definitely keep this conversation going beyond this PR!

TomWhite-MedStar:
@katy-sadowski , the only conflict is in listChecks.R.  I submitted a separate PR for that file #452, since it should have been part of PR #444 (sorry for missing that).

I'll review the other changes.  I just need to remind myself of prior functionality.  I believe the only difference was to take the main SQL query output and insert it into the cte (so that it can be inserted into the dqd results table).  I can make that step optional.

I don't use the Shiny app.  Instead, I write the data to a database, and we use Tableau to review overall status.  That way we can also review and filter by number of records in the denominator and numerator - plus the thresholds that were used in the logic. In some cases, our % fail is high, but the denominator count is so small that the impact is immaterial.

TomWhite-MedStar:
@katy-sadowski and @MaximMoinat , I made the requested change so that sqlOnly can be backwards compatible.  In order to use the new functionality, uses must now also set sqlOnlyIncrementaInsert <- TRUE.  I updated the documentation to reflect that.

Once PR #452 is accepted, this PR should merge without conflicts.

R/sqlOnly.R
TomWhite-MedStar: done

R/readThresholdFile.R
```diff
@@ -15,20 +15,23 @@
...
+  colspec <- readr::spec_csv(thresholdFile)
+  # plausibleUnitConceptIds is a comma-separated list of concept ids, but it is being interpreted as col_double()
+  if ('plausibleUnitConceptIds' %in% names(colspec$cols)) {
+    colspec$cols$plausibleUnitConceptIds <- readr::col_character()
+  }
+  result <- read_csv(
+    file = thresholdFile,
+    col_types = colspec,
+    na = c(" ", "")
+  )
```
katy-sadowski, line 35: for some reason the diff here isn't showing [this](https://github.com/OHDSI/DataQualityDashboard/blob/eb8a749b6a6ffeeca1c8c2ab68f62a90d03835f6/R/readThresholdFile.R#L40) change i added in v2.2. can you try rebasing again off the develop branch? i think doing that should pull in the right state of develop here.

katy-sadowski (commented):
Thanks for the updates @TomWhite-MedStar !  Just 2 more small changes requested.

I'm doing some docs updates to emphasize the "beta" status and to make sure everything is super clear for users, and am seeing if I can add a unit test that runs a UNION'ed query or 2 (I will push these changes up to your branch if that's OK!).  And then I think we should be good to go!  There are a couple other minor changes I'd like to squeeze in the next release but we should be on track to ship it by the end of next week if all goes well :)

TomWhite-MedStar:
@katy-sadowski , made all requested changes.

TomWhite-MedStar:
@katy-sadowski , I added a unit test to testthat to confirm that sqlOnlyUnionCount and sqlOnlyIncrementalInsert work properly.  It includes a check of whether the generated SQL matches expectations.  If that approach seems reasonable, it would be easy to add several more with different permutations of the parameters - including testing against the legacy format.

BTW, R made it much harder than necessary to do the regex substitutions needed to remove single line and multi-line comments, but the new remove_sql_comments() function seems to work properly.  It might also help with SqlRender testing, as it doesn't have an easy way to deal with multi-line output yet.

katy-sadowski:
Thanks @TomWhite-MedStar, and apologies for the delay.  I was out of town at the end of last week.  I just opened a PR in your repo with some refactors - please take a look and let me know what you think.  The PR is against your fix_sqlOnly branch so once merged in my commits should appear here :) 

I'm now just working on some updates to the sqlOnly vignette.  Once we agree on my proposed changes and I update the docs this should be good to merge in!

TomWhite-MedStar:
@katy-sadowski , thanks for making those changes!

katy-sadowski:
@TomWhite-MedStar @MaximMoinat I added 1 more PR with some updates to the vignette: https://github.com/TomWhite-MedStar/DataQualityDashboard/pull/6.  Please take a look - I think that should be the final step before merging this PR into develop and then prepping the release!

TomWhite-MedStar:
@katy-sadowski , thanks for making the changes to the vignettes.

katy-sadowski:
@TomWhite-MedStar I was doing a final review before merging and noticed a few more small things that needed updating in the docs. PR here: https://github.com/TomWhite-MedStar/DataQualityDashboard/pull/7. After this one we will really really be done! Thanks!

TomWhite-MedStar:
@katy-sadowski , thanks for making those changes.  Merged them.

katy-sadowski (approved):

PR #444, opened 2023-04-30
Fix issue #436 - plausibleUnitConceptIds not read correctly from csv

TomWhite-MedStar:
Fix issue #436. 

Root cause was that plausibleUnitConceptIds were being read as numbers via read_csv() instead of strings.

### DIFF
--- a/R/readThresholdFile.R
+++ b/R/readThresholdFile.R
@@ -15,20 +15,23 @@
 # limitations under the License.
 
 .readThresholdFile <- function(checkThresholdLoc, defaultLoc) {
+  thresholdFile = checkThresholdLoc
   if (checkThresholdLoc == "default") {
-    result <- read_csv(
-      file = system.file(
-        "csv",
-        defaultLoc,
-        package = "DataQualityDashboard"
-      ),
-      na = c(" ", "")
-    )
-  } else {
-    result <- read_csv(
-      file = checkThresholdLoc,
-      na = c(" ", "")
+    thresholdFile = system.file(
+      "csv",
+      defaultLoc,
+      package = "DataQualityDashboard"
     )
   }
+  colspec = spec_csv(thresholdFile)
+  # plausibleUnitConceptIds is a comma-separated list of concept ids, but it is being interpreted as col_double()
+  if ('plausibleUnitConceptIds' %in% names(colspec$cols)) {
+    colspec$cols$plausibleUnitConceptIds = col_character()
+  }
+  result <- read_csv(
+    file = thresholdFile,
+    col_types = colspec,
+    na = c(" ", "")
+  )
   return(result)
 }
--- a/R/readThresholdFile.R
+++ b/R/readThresholdFile.R
@@ -15,18 +15,18 @@
 # limitations under the License.
 
 .readThresholdFile <- function(checkThresholdLoc, defaultLoc) {
-  thresholdFile = checkThresholdLoc
+  thresholdFile <- checkThresholdLoc
   if (checkThresholdLoc == "default") {
-    thresholdFile = system.file(
+    thresholdFile <- system.file(
       "csv",
       defaultLoc,
       package = "DataQualityDashboard"
     )
   }
-  colspec = spec_csv(thresholdFile)
+  colspec <- readr::spec_csv(thresholdFile)
   # plausibleUnitConceptIds is a comma-separated list of concept ids, but it is being interpreted as col_double()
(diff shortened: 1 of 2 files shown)


R/readThresholdFile.R
```diff
@@ -15,20 +15,23 @@
 # limitations under the License.
 
 .readThresholdFile <- function(checkThresholdLoc, defaultLoc) {
+  thresholdFile = checkThresholdLoc
```
MaximMoinat, line 18:
```suggestion
  thresholdFile <- checkThresholdLoc
```

R/readThresholdFile.R
```diff
@@ -15,20 +15,23 @@
...
-        defaultLoc,
-        package = "DataQualityDashboard"
-      ),
-      na = c(" ", "")
-    )
-  } else {
-    result <- read_csv(
-      file = checkThresholdLoc,
-      na = c(" ", "")
+    thresholdFile = system.file(
```
MaximMoinat, line 20:
```suggestion
    thresholdFile <- system.file(
```

R/readThresholdFile.R
```diff
@@ -15,20 +15,23 @@
...
+      "csv",
+      defaultLoc,
+      package = "DataQualityDashboard"
     )
   }
+  colspec = spec_csv(thresholdFile)
```
MaximMoinat, line 26:
```suggestion
  colspec <- readr::spec_csv(thresholdFile)
```

R/readThresholdFile.R
```diff
@@ -15,20 +15,23 @@
...
+  # plausibleUnitConceptIds is a comma-separated list of concept ids, but it is being interpreted as col_double()
+  if ('plausibleUnitConceptIds' %in% names(colspec$cols)) {
+    colspec$cols$plausibleUnitConceptIds = col_character()
```
MaximMoinat, line 29:
```suggestion
    colspec$cols$plausibleUnitConceptIds <- col_character()
```

MaximMoinat (approved):
Works for me. Please review and accept the suggested style changes

TomWhite-MedStar:
@MaximMoinat , thanks.  I accepted the style changes, plus had to explicitly call readr::col_character() rather than col_character() to avoid a run-time error.

TomWhite-MedStar:
@MaximMoinat , Although this PR fixes the bug, I'm noticing two more things:
1. I don't think the logic is working as intended for situations where no list of plausibleUnitConceptIds is specified.  In those cases, you get a query like this for violated rows:

```
 WHERE m.MEASUREMENT_CONCEPT_ID = 3015274
 AND 
 m.unit_concept_id IS NOT NULL
```

In situations where plausibleUnitConceptIds  is NA, I presume that no DQD check should be generated.  As is, this will conflict with other checks about the % of units with NULL value.

2. There is no vignette for plausibleUnitConceptIds.  I drafted one, but not sure whether to add that to this PR or elsewhere.

TomWhite-MedStar:
1. Added a vignette for plausibleUnitConceptIds

2. The SQL is working as designed in [concept_plausible_unit_concept_ids.sql](https://github.com/OHDSI/DataQualityDashboard/blob/main/inst/sql/sql_server/concept_plausible_unit_concept_ids.sql), but it may not be working as intended.  The relevant section is lines 40-45:

```
		  AND {@plausibleUnitConceptIds == '' | @plausibleUnitConceptIds == 'NA'}?{
		  m.unit_concept_id IS NOT NULL
		  }:{
		  m.unit_concept_id NOT IN (@plausibleUnitConceptIds)
		  }
```

katy-sadowski:
Thanks @TomWhite-MedStar and @MaximMoinat for this fix and for updating the vignette!  I also tested it locally and it looks good to me.

Regarding handling of concepts without `plausibleUnitConceptIds` specified - the check will only run if a threshold is specified (see [here](https://github.com/OHDSI/DataQualityDashboard/blob/main/inst/csv/OMOP_CDMv5.3_Check_Descriptions.csv#L23)).  I think the idea is that if you specify a threshold but no unit concepts, the concept in question should not have any units.  I checked a couple of these rows in the threshold files and this checks out; the concepts with a threshold but without plausible unit concepts are "presence" measurements rather than continuous numeric measurements with a unit.

katy-sadowski:
@TomWhite-MedStar I'm going to go ahead and merge in this change so I can release it with v2.2 which is about to go out. Let me know if you have further questions/input on plausibleUnitConceptIds.

PR #447, opened 2023-05-01
Update codeToRun.R

MaximMoinat:
No functional changes, just fixing the example code.

Besides bit of clean-up, the main fix is that two variables were assigned but not used as a parameter (`cdmVersion`  and `outputFile`). These are not added as parameter of `executeDqChecks`.


katy-sadowski:
LGTM!  Can you please update target branch to `develop`?  Then we can merge in 👌

MaximMoinat:
Thanks @katy-sadowski, I switched the branch just now!

katy-sadowski (approved):

Issue #451, opened 2023-05-06
listChecks.R does not read plausibleConceptIds correctly


TomWhite-MedStar:
It reads them as double instead of a string of comma-separated values

TomWhite-MedStar:
This issue was fixed by PR #452.

PR #452, opened 2023-05-06
Fix issue #451 - properly read plausibleUnitConceptIds

TomWhite-MedStar:
Fix issue #451 

.readThresholdFile() is used everywhere else.  This makes listChecks.R consistent with that approach.

This is also related to issue #436 and pull request #444


katy-sadowski (approved):
Thanks for this optimization!

Issue #445, opened 2023-05-01
Rewrite `temporalPlausibleAfter` for checking events before birth


MaximMoinat:
The check `temporalPlausibleAfter` is used to check for dates before birth based on `PERSON.BIRTH_DATETIME`.
However, this is a non-required field and therefore will be empty in many databases.

I would propose to use `PERSON.YEAR_OF_BIRTH` by default, for which there is already logic in [field_plausible_temporal_after.sql](https://github.com/OHDSI/DataQualityDashboard/blob/1197d5d8287c35a5c3f71d010220959494a160d3/inst/sql/sql_server/field_plausible_temporal_after.sql#L45-L46). To make this work, the `plausibleTemporalAfterTableName` is set to PERSON and `plausibleTemporalAfterFieldName` has to be left empty in the field-level thresholds file.

MaximMoinat:
Another note: the logic uses June 1st is used as a proxy together with the year of birth. This is tricky, as half of the persons will be born before that.

We could combine all four date of birth fields into one query, using the information that is available and defaulting to the first day of the month/year when unavailable.

```sql
COALESCE(
    plausibleTable.birth_datetime, 
    CAST(CONCAT(
        plausibleTable.year_of_birth, '-',
        COALESCE(plausibleTable.month_of_birth, 1), '-',
        COALESCE(plausibleTable.day_of_birth, 1)
     ) AS DATE)
)
```

To replace the coalesce in:
https://github.com/OHDSI/DataQualityDashboard/blob/1197d5d8287c35a5c3f71d010220959494a160d3/inst/sql/sql_server/field_plausible_temporal_after.sql#L43-L48

katy-sadowski:
I agree with the proposal here!  Also agree that comparisons to DOB may be better off in their own check to reduce complexity of `temporalPlausibleAfter` ... I was just thinking we could add these into `plausibleDuringLife` instead and then realized that's maybe what you're up to in https://github.com/OHDSI/DataQualityDashboard/pull/454?  If yes - I support it!

MaximMoinat:
Thanks Katy. I will indeed add this to the PR you mentioned, refactoring the `plausibleDuringLife` check.

PR #454, opened 2023-05-20
Redefine field_plausible_during_life and field_plausible_temporal_after

MaximMoinat:
Redefine plausibleDuringLife, to be in line with the name; checking for events that happen before birth OR after death, instead of only checking for events after death.

Removes the date of birth check from [field_plausible_temporal_after.sql](https://github.com/OHDSI/DataQualityDashboard/blob/1197d5d8287c35a5c3f71d010220959494a160d3/inst/sql/sql_server/field_plausible_temporal_after.sql#L45-L46).

Lastly, the date of birth derivation logic has changed. It is taken either from birth_datetime if available, otherwise calculated from year/month/day_of_birth, taking 1st if month/day not available (instead of using July 1st). See #445.


MaximMoinat:
To the reviewer(s): this check changes the meaning of both checks, making results harder to compare between DQD versions. It might be better to create completely new checks.

dimshitc (commented):
All good, but we agreed to split it into 3 checks. So, please let me know when this is done, and I'll approve the PR.

MaximMoinat:
Closing this PR, reimplemented as three new checks in #516

Issue #484, opened 2023-09-12
Please update unit tests to use CDM54 schema


schuemie:
See HADES Teams for details

katy-sadowski:
Thanks for the alert!  This is now updated in develop.  Will ship a new release soon including this change.

katy-sadowski:
Addressed in v2.4.1 release.

PR #485, opened 2023-09-14
Update Weekly R Check Workflow to Use CDM v5.4; Add Redshift Tests

katy-sadowski:
Stuff I missed in #479. Addresses #484.


katy-sadowski:
Forgot to add download of Redshift driver in test setup; this has now been added in #486

PR #253, opened 2021-09-20
New way to edit thresholds

MaximMoinat:
Adds function `editThresholds(threshold_file)` that can create a thresholds file from a table of modified thresholds.

As described under point 1 in this OHDSI 2021 collaborator showcase: https://www.ohdsi.org/2021-global-symposium-showcase-67/

Making it a draft as it needs to be moved from `extras` to the main codebase, with Roxygen doc. We do include a readme here. Any feedback at this stage is much appreciated.

PR #516, opened 2023-12-21
Add three new temporal checks: `plausibleAfterBirth`, `plausibleBeforeDeath`, `plausibleStartBeforeEnd`

### LINKED ISSUE #445: Rewrite `temporalPlausibleAfter` for checking events before birth

### LINKED ISSUE #513: Modify the check: The number and percent of records with a value in the OBSERVATION_PERIOD_START_DATE field of the OBSERVATION_PERIOD that occurs prior to the date in the BIRTH_DATETIME field of the PERSON table.
Need to modify the check. Currently if the only year of birth is given, it uses CAST(CONCAT(plausibleTable.year_of_birth,'-06-01') AS DATE. But the real date of birth can be easily in the beginning of the year, leading to observation period start at this date.

MaximMoinat:
As discussed in #454: these checks separate out the quality checks in plausibleDuringLife and plausibleTemportalAfter.
- `plausibleAfterBirth` checks for any events happening before birth (was implicitly checked in plausibleTemportalAfter)
- `plausibleBeforeDeath` checks for any events happening after death (improving upon implementation of plausibleDuringLife)
- `plausibleStartBeforeEnd` checks for end dates that are before start date in the same table (specific implementation of plausibleTemportalAfter)

TODO:
- [x] Add threshold settings for v5.2 and v5.3 (after validating that this works for v5.4)
- [x] To discuss: disable `plausibleDuringLife` and `plausibleTemportalAfter` by default. These are redundant if the three new checks are run, but kept in for backwards compatibility.

fixes #513 and fixes #445


dimshitc (commented):
why do we keep plausibleTemporalAfter and plausibleDuringLife checks?
aren't they covered now by these new 3 checks?

MaximMoinat:
> why do we keep plausibleTemporalAfter and plausibleDuringLife checks? aren't they covered now by these new 3 checks?

Removing the 'replaced' checks might be a breaking change for some people (e.g. when comparing number of violating records over time). My thinking was to implement a 'soft' disable, making it possible to execute the old checks if desired. e.g. by removing the plausibleDuringLife and plausibleTemportalAfter checks from the `checkDescriptions` object [here](https://github.com/OHDSI/DataQualityDashboard/blob/81230bde213b99c5d26a177c7922b70a15a06349/R/executeDqChecks.R#L253), and only keep them if specified specifically in the `checkNames` parameter.

R/executeDqChecks.R
```diff
@@ -252,6 +253,10 @@ executeDqChecks <- function(connectionDetails,
 
   if (length(checkNames) > 0) {
     checkDescriptionsDf <- checkDescriptionsDf[checkDescriptionsDf$checkName %in% checkNames, ]
+  } else {
+    # No checkNames specified. Do not run plausibleDuringLife and plausibleTemportalAfter as these are superseded three new checks
```
MaximMoinat, line 257: @dimshitc This disables the old checks, if user did not specify them specifically. We do need to document this behaviour.

dimshitc: This makes sense to me.

Should we set threshold in all these checks to 0? as these are obvious mistakes and shouldn't be in the data.

MaximMoinat: Yes, that makes even more sense. I added the following warnings if one of the old checks is run:
`"DEPRECATION WARNING - The plausibleDuringLife check has been reimplemented with the plausibleBeforeDeath check."`

`DEPRECATION WARNING - The plausibleTemporalAfter check has been reimplemented with the plausibleAfterBirth and plausibleStartBeforeEnd checks.`

MaximMoinat: Regarding disabling the checks by the user; there is not really a nice solution as `executeDqChecks` only takes the names of checks to be run, not an argument with checks to exclude. So the user needs to list all checks except these two. Or does one of you see a more elegant solution?

inst/sql/sql_server/field_plausible_after_birth.sql
```diff
@@ -0,0 +1,62 @@
...
+        JOIN @cohortDatabaseSchema.@cohortTableName c ON cdmTable.person_id = c.subject_id
+            AND c.COHORT_DEFINITION_ID = @cohortDefinitionId
+        }
+        JOIN @cdmDatabaseSchema.person p ON cdmTable.person_id = p.person_id
+        WHERE CAST(cdmTable.@cdmFieldName AS DATE) < COALESCE(
+                p.birth_datetime, 
+                CAST(CONCAT(
+                    p.year_of_birth, '-',
+                    COALESCE(p.month_of_birth, 1), '-',
+                    COALESCE(p.day_of_birth, 1)
```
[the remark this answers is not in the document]
MaximMoinat: Good catch, I actually came across [a similar issues in sqlite](https://github.com/OHDSI/SqlRender/issues/351). 

Not a trivial change, but I came up with the following code, which is a bit complex but should be compatible with all Sql dialects:
```sql
CAST(CONCAT(
    p.year_of_birth,
    COALESCE(
        RIGHT('00' + CAST(p.month_of_birth AS VARCHAR), 2),
        '01'
    ),
    COALESCE(
        RIGHT('00' + CAST(p.day_of_birth AS VARCHAR), 2),
        '01'
    )
) AS DATE)
```

inst/sql/sql_server/field_plausible_before_death.sql
```diff
@@ -0,0 +1,57 @@
...
+        SELECT 
+            '@cdmTableName.@cdmFieldName' AS violating_field, 
+            cdmTable.*
+        FROM @cdmDatabaseSchema.@cdmTableName cdmTable
+        {@cohort & '@runForCohort' == 'Yes'} ? {
+        JOIN @cohortDatabaseSchema.@cohortTableName c ON cdmTable.person_id = c.subject_id
+            AND c.COHORT_DEFINITION_ID = @cohortDefinitionId
+        }
+        JOIN @cdmDatabaseSchema.death de ON cdmTable.person_id = de.person_id
+        WHERE cast(cdmTable.@cdmFieldName AS DATE) > DATEADD(day, 60, cast(de.death_date AS DATE))
```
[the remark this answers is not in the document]
MaximMoinat: Done

inst/sql/sql_server/field_plausible_after_birth.sql
```diff
@@ -0,0 +1,62 @@
...
+        SELECT 
+            '@cdmTableName.@cdmFieldName' AS violating_field, 
+            cdmTable.*
+        FROM @cdmDatabaseSchema.@cdmTableName cdmTable
+        {@cohort & '@runForCohort' == 'Yes'} ? {
+        JOIN @cohortDatabaseSchema.@cohortTableName c ON cdmTable.person_id = c.subject_id
+            AND c.COHORT_DEFINITION_ID = @cohortDefinitionId
+        }
+        JOIN @cdmDatabaseSchema.person p ON cdmTable.person_id = p.person_id
+        WHERE CAST(cdmTable.@cdmFieldName AS DATE) < COALESCE(
```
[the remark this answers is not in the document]
MaximMoinat: Agree, added same condition for the other two checks as well

R/executeDqChecks.R
katy-sadowski: Warnings look great; thanks!  

And I wouldn't call it elegant, but I did add these instructions to the docs earlier this year on how to exclude a list of checks: https://github.com/OHDSI/DataQualityDashboard/blob/81230bde213b99c5d26a177c7922b70a15a06349/vignettes/DataQualityDashboard.rmd#L114-L120

I think for now we can point to this in the release notes.  In the future I'd love to improve the workflow for enabling/disabling checks and editing thresholds (a la #253 ).

inst/sql/sql_server/field_plausible_after_birth.sql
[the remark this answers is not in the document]
katy-sadowski: hah, crafty little solution!  only question - do we need `00` or just `0` there? (won't actually impact the result)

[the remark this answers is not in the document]
MaximMoinat: Well spotted, will change it

inst/sql/sql_server/field_plausible_before_death.sql
```diff
@@ -1,7 +1,7 @@
 
 /*********
 PLAUSIBLE_BEFORE_DEATH
-all events occur after death (PLAUSIBLE_BEFORE_DEATH == Yes)
+Chedcks for events that occur more than 60 days after death (PLAUSIBLE_BEFORE_DEATH == Yes).
```
katy-sadowski, line 4: small typo here!

MaximMoinat, line 4:
```suggestion
Checks for events that occur more than 60 days after death (PLAUSIBLE_BEFORE_DEATH == Yes).
```

inst/sql/sql_server/field_plausible_after_birth.sql
```diff
@@ -0,0 +1,70 @@
+/*********
+PLAUSIBLE_AFTER_BIRTH
+all events happen after birth (PLAUSIBLE_AFTER_BIRTH == Yes)
```
katy-sadowski, line 3: Add "Checks that..."

inst/sql/sql_server/field_plausible_start_before_end.sql
```diff
@@ -0,0 +1,61 @@
+
+/*********
+PLAUSIBLE_START_BEFORE_END
+all start dates are before their corresponding end dates (PLAUSIBLE_START_BEFORE_END == Yes).
```
katy-sadowski, line 4: Add "Checks that..."

inst/sql/sql_server/field_plausible_after_birth.sql
```diff
@@ -0,0 +1,70 @@
...
+Birthdate is either birth_datetime or composed from year_of_birth, month_of_birth, day_of_birth (taking 1st month/1st day if missing).
+Denominator is number of events.
```
katy-sadowski, line 5: Add "...with a non-null date"

inst/sql/sql_server/field_plausible_before_death.sql
```diff
@@ -0,0 +1,59 @@
+
+/*********
+PLAUSIBLE_BEFORE_DEATH
+Checks for events that occur more than 60 days after death (PLAUSIBLE_BEFORE_DEATH == Yes).
+Denominator is number of events of persons who died.
```
katy-sadowski, line 5: "number of events with a non-null date"

inst/sql/sql_server/field_plausible_start_before_end.sql
```diff
@@ -0,0 +1,61 @@
...
+@cdmFieldName is the start date and @plausibleStartBeforeEndFieldName is the end date.
+
+Parameters used in this template:
+schema = @schema
+cdmTableName = @cdmTableName
+cdmFieldName = @cdmFieldName
+plausibleTemporalAfterTableName = @plausibleTemporalAfterTableName
+plausibleTemporalAfterFieldName = @plausibleTemporalAfterFieldName
```
katy-sadowski, line 12: lines 11 and 12 are duplicated, and should reference `@plausibleStartBeforeEndFieldName` instead

katy-sadowski (changes requested):
OK I promise this is my last review 😅  I did a deep review and left a couple more comments on the SQL comments.  I also noticed the following in the threshold files (I only checked 5.4 but assume they're all the same):

- EPISODE start_date and start_datetime are missing a threshold for plausibleStartBeforeEnd
- The plausibleStartBeforeEnd threshold for OBSERVATION_PERIOD is 1.  Should this actually be 0, since this is a derived table and folks can control it in ETL (and also would prob cause some pretty weird stuff in analytics if the dates were flipped)?
- plausibleAfterBirth and plausibleBeforeDeath need to be activated for MEASUREMENT.measurement_datetime

MaximMoinat:
Thanks @katy-sadowski for this thorough review, I agree with all your comments and adjusted the code accordingly.

katy-sadowski (approved):
Looks great! Thanks so much @MaximMoinat! I will go ahead and merge this into develop.

PR #519, opened 2024-01-15
check gender specific events - descendants of a given concept

dimshitc:
This adds new check when a concept set of gender specific concepts is evaluated - whether patients with these events have a correct gender.
Changes were made only in CDM v54, if approved, I'll add them to the other CDM versions.

image

Currently this check looks similar to the checks without descendant usage (In the image above I had to filter by description including 'descendants'). Please let me know if we need to distinguish it more.


katy-sadowski (commented):
Thanks for this @dimshitc !  Overall looks good but I have a couple of questions:

- What are your thoughts on flipping this around so that the denominator is # persons with the gender, and numerator is # persons w/ an implausible concept for that gender?  I *think* that doing so would make more sense from a database-wide data quality perspective.  But need to ponder this more - this could be a good one to discuss again at our next DQ meeting
- Do the ancestor concepts we use here subsume all of the other individual concepts in the threshold file?  If yes - I'm thinking that we could deprecate the old plausibleGender checks the same way Maxim did for temporal plausibility checks here: https://github.com/OHDSI/DataQualityDashboard/pull/516/files# (with the idea to eventually get rid of them altogether).  If they don't subsume all of the other concepts, maybe we could add a couple more ancestor rows to get the rest of them?
- I noticed we have Procedure on female genital system ancestor but not the same for males, should we add it?

Final thing - please change the target branch to `develop`.  (You may need to rebase your branch if you pulled from main originally).

dimshitc:
Thanks Katy!

1) it seems to be incomparable cohorts, for example in CCAE, only 12% of male patients have any _Male genitalia finding_ condition. And yes, let's discuss it on the call.

2.1. conditions
males codes to be added
 4025213 -- 'male reproductive finding' to be added to male descendants
female codes to be added
443343 -- Disorder of female genital system
4024004	Female reproductive finding
4172857 [Menopause finding](https://athena.ohdsi.org/search-terms/terms/4172857)
444094 Finding related to pregnancy - risk of including codes related to the newborn
197810, 4158481
i.e. all female codes:
( 4095793 , 443343, 4024004 , 4172857, 444094 , 197810, 4158481)

3. There are wrong hierarchy for male procedures, and I don't know when it will be fixed, so better not to include it now.

dimshitc:
Ok, when I rebase to the develop it shows 37 files changed, so it becomes hard to track changes, I suggest we agree on all the changes, and then I create the new branch from the develop, with the latest changes

katy-sadowski:
@dimshitc Sounds good for those additional concepts, and for leaving off male procedures for now.  For 444094 I did a quick check of the ICD10-CMs and there is only one P-code (perinatal record code) in there, which is probably a mis-mapping?  Not sure about other vocabs, but assume/hope that this indicates 444094 should only include findings on the mother?

Also agree for starting a fresh branch off develop rather than dealing with the main<>develop diff.  Maybe you can put the changes from your other new check PR on there too, and we can merge them in together? :) 

Look forward to chatting about the denominator in the meeting Thurs.  It's unideal both ways, but now I see how it could end up being a really small failure % and hard to make a threshold for, if we make the denominator persons.

PR #527, opened 2024-02-13
NotApplicable fix for measureConditionEraCompleteness

### LINKED ISSUE #526: measureConditionEraCompleteness is 'Not Applicable' if condition era table is empty
## Expected behaviour
The check `measureConditionEraCompleteness` checks if condition era is populated and should fail if not all persons that have a condition occurrence also have a condition era.

## Actual behaviour
If Condition Era table is empty, the check status will be 'Not Applicable' (even though it correctly determines that 100% of records fail).

image

And reversely, if Condition Occurrence is empty and Condition Era populated, `measureConditionEraCompleteness` will always pass. Here it should be NA as the check has become meaningless.

## Suggested fix
The `measureConditionEraCompleteness` check status should be NA if the condition_occurrence is empty. To do this, I see two options: 1) make a separate variable against which table/field should be checked for emptyness or 2) make an exception for this check.

MaximMoinat:
Fixes #526, by having an exception for measureConditionEraCompleteness check; it will be NA when condition occurrence table is missing or empty.

To be able to make the fix, I had to refactor the NA status code. While doing that, I also moved the logic to a separate function.

As a reminder when/if we add measureDrugEraCompleteness, I have added some code for that already but commented it out for now.

TODO:
- [x] Investigate failing unit tests
- [x] Add tests for `.calculateNAstatus`


MaximMoinat:
I see the checks ar failing, I will investigate and fix this.

katy-sadowski:
Thanks @MaximMoinat !  Love the initiative.  I'll wait to do a deep review til you've sorted out the failing checks.  I'd also recommend to add some unit tests for the code.  I know there weren't any test specific to NA before, but now that they're in their own function and we're refactoring a bit, it feels like the right time.

Might need to get a bit crafty to test this though :) off the top of my head, I'm picturing truncating a table in Eunomia and then seeing if related checks are marked NA. Happy to help brainstorm further on that.

MaximMoinat:
I like the idea of adding tests, will look into that.

> Might need to get a bit crafty to test this though :) off the top of my head, I'm picturing truncating a table in Eunomia and then seeing if related checks are marked NA. Happy to help brainstorm further on that.

That is exactly the test I did locally already: run with truncated condition occurrence (expecting NA) and with truncated condition era (expecting fail). This passed for me, but would indeed be good to have this integrated into the test suite.

MaximMoinat:
@katy-sadowski Tests are fixed and new tests added. Note that these are in a separate file "test-calculateNotApplicableStatus.R", although they do not test the `.calculateNotApplicableStatus` function in isolation, but by running the full `executeDqChecks`.

R/calculateNotApplicableStatus.R
```diff
@@ -0,0 +1,188 @@
...
+      ),
+      failed = ifelse(.data$notApplicable == 1, 0, .data$failed),
+      passed = ifelse(.data$failed == 0 & .data$isError == 0 & .data$notApplicable == 0, 1, 0)
+    ) %>%
+    dplyr::select(-c("tableIsMissing", "fieldIsMissing", "tableIsEmpty", "fieldIsEmpty", "conceptIsMissing", "conceptAndUnitAreMissing"))
+
+  return(checkResults)
+}
+
+.applyNotApplicable <- function(x) {
```
katy-sadowski, line 163: could this go above the main function for readability / my preference? :)

R/calculateNotApplicableStatus.R
```diff
@@ -0,0 +1,188 @@
...
+#' Determines if all checks required for 'Not Applicable' status are in the checkNames
+#'
+#' @param checkNames A character vector of check names
+#'
+#' @keywords internal
+.containsNAchecks <- function(checkNames) {
+  naCheckNames <- c("cdmTable", "cdmField", "measureValueCompleteness")
+  missingNAChecks <- !(naCheckNames %in% checkNames)
+  if (any(missingNAChecks)) {
+    missingNACheckNames <- paste(naCheckNames[missingNAChecks], collapse = ", ")
```
katy-sadowski, line 36: this isn't used anywhere, do you want to make that the return value instead?

tests/testthat/test-calculateNotApplicableStatus.R
```diff
@@ -0,0 +1,87 @@
...
+    resultsDatabaseSchema = resultsDatabaseSchemaEunomia,
+    cdmSourceName = "Eunomia",
+    checkNames = c("cdmTable", "cdmField", "measureValueCompleteness"),
+    # Eunomia COST table has misspelled 'REVEUE_CODE_SOURCE_VALUE'
+    tablesToExclude = c("COST", "CONCEPT", "VOCABULARY", "CONCEPT_ANCESTOR", "CONCEPT_RELATIONSHIP", "CONCEPT_CLASS", "CONCEPT_SYNONYM", "RELATIONSHIP", "DOMAIN"),
+    outputFolder = outputFolder,
+    writeToTable = F
+  )
+
+  # Assumption that Eunomia has empty device_exposure table
```
katy-sadowski, line 19: a bold move ;) if eunomia updates to fill that table, then our test starts failing. maybe clear it out like you did below, to be safe?

tests/testthat/test-calculateNotApplicableStatus.R
```diff
@@ -0,0 +1,87 @@
...
+                            results$CheckResults$tableName == "device_exposure", ]
+  expect_true(all(r$notApplicable == 1))
+})
+
+test_that("measureConditionEraCompleteness Not Applicable if condition_occurrence empty", {
+  outputFolder <- tempfile("dqd_")
+  on.exit(unlink(outputFolder, recursive = TRUE))
+
+  # Remove records from Condition Occurrence
+  connection <- DatabaseConnector::connect(connectionDetailsEunomia)
```
katy-sadowski, line 30: i think it would be safer to create a separate eunomia DB just for these tests. i believe testthat runs the test files in parallel so we might run into weird issues if we're modifying the same DB here.

katy-sadowski (changes requested):
SORRY for the delay!!!  This looks good to me.  Just a couple minor comments.  Thanks so much for adding this!

MaximMoinat:
Thanks for your comments Katy. I agree with your suggestions and will implement these.

tests/testthat/test-calculateNotApplicableStatus.R
MaximMoinat: I don't think this is an issue. If I open a Eunomia connection in two R sessions, clearing a table in one session does not affect the other session.

MaximMoinat:
@katy-sadowski Implemented the requested changes, except for the one on the test. See my comment.

tests/testthat/test-calculateNotApplicableStatus.R
katy-sadowski: @MaximMoinat I think this got lost in the shuffle - let me know what you think and if you're OK creating new connection details for these tests. Thanks!

tests/testthat/test-calculateNotApplicableStatus.R
```diff
@@ -4,19 +4,23 @@ test_that("Not Applicable status Table Empty", {
   outputFolder <- tempfile("dqd_")
   on.exit(unlink(outputFolder, recursive = TRUE))
 
+  # Make sure the device exposure table is empty
+  connection <- DatabaseConnector::connect(connectionDetailsEunomia)
```
katy-sadowski, line 8: should this be connectionDetailsEunomiaNaChecks?

MaximMoinat: Yes, thanks, fixed now.

katy-sadowski (approved):

Issue #500, opened 2023-10-21
Update check type descriptions & add user guidance


katy-sadowski:
Update the following docs with updated descriptions and new user guidance from [DQD Hackathon](https://ohdsiorg.sharepoint.com/:x:/s/Workgroup-Dataquality/EXF1gbCJkxhNtovc5UtYgwkBw2w1Wtmv5Kh1G78JAPqk7A?e=VwXtjg):

- https://github.com/OHDSI/DataQualityDashboard/blob/main/vignettes/CheckTypeDescriptions.rmd
- https://github.com/OHDSI/DataQualityDashboard/blob/main/inst/csv/OMOP_CDMv5.4_Check_Descriptions.csv (as needed)
- https://github.com/OHDSI/DataQualityDashboard/tree/main/inst/sql/sql_server (query headers, as needed)

Issue #323, opened 2022-05-03
New Procedure Checks


clairblacketer:
Some ideas for new checks on PROCEDURE_OCCURRENCE

procedures that last a long time > 3 days?

multiple procedures that can only happen once (kidney removal, replacement of the same joint in a certain amount of time)

Issue #289, opened 2022-03-10
Inpatient visits that only last a day


clairblacketer:
If an inpatient visit only lasts one day then they should not be classified as inpatient, this would be a new check

This could also be applied to other visits, outpatient should only be 1 day, ER should only be 1 day, <24hrs

MelaniePhilofsky:
Let's not add "ER should only be 1 day, <24hrs". In general, this should be true. But in periods of high demand and low available bed counts, some persons stay in the ER for > 24 hours. 48 hours would be a more helpful check.

clairblacketer:
@MelaniePhilofsky thank you for the insight!

priagopal:
Hi, I can try to create this check in SQL. I wasn't sure, if there is a way for me to assign the issue to myself?

Thanks,
Priya

clairblacketer:
Thanks @priagopal! I added you as a collaborator and assigned the issue to you

priagopal:
Thank you

priagopal:
Hi Clair,
I have created the check. Can you please check? if it looks good. I can take more issues related to SQL checks.

Thanks,
Priya

katy-sadowski:
hi @clairblacketer & @priagopal , i saw that the PR for this check in the develop branch ended up getting reverted (#336 ). is the intention to develop/add this check later on in the future, or not to add after all?

PR #534, opened 2024-03-14
Add logic to support context-specific modules

jshoughtaling:
This PR proposes a new architecture that allows users to (1) apply a targeted subset of checks based on a particular context of interest (e.g. ICU, pregnancy, oncology, etc.), and (2) add modules that include multiple complex checks at any level (TABLE, FIELD, CONCEPT) without needing to add multiple columns per check to the orchestration csv files. I've also incorporated a first demo module (ICU) that can be selected by passing a single argument (`contextModule` = `ICU`) to the main `executeDqChecks` function. 

It also adds documentation about how to create and insert new modules, either for local use or for dissemination within focused consortia. 

The code changes address the following issues:
- https://github.com/OHDSI/DataQualityDashboard/issues/500
- https://github.com/OHDSI/DataQualityDashboard/issues/323
- https://github.com/OHDSI/DataQualityDashboard/issues/289

Possible elements to discuss:
- Ability to pull checks/modules from separate repositories
- Ability to incorporate pre-defined concept sets (e.g. TermHub) into modules
- Separation of module logic (sql) and orchestration (csv) into sub directories or other locations

High-level view of modular structure:

DQDModules drawio (1)


katy-sadowski:
Hi @jshoughtaling thanks again for working on this!  One thing I'd like to add which I don't think was mentioned in our call the other week - one requirement I think should be imposed for a custom-checks solution is that users should be able to add their own checks _without_ modifying the package itself.  So the custom SQL queries should live in a directory outside DQD which DQD can be pointed at as the location for the custom SQL.

Let me know what you think about trying to incorporate this here :)

PR #554, opened 2024-06-26
Add more docs pages

katy-sadowski:
Adds:

- measureValueCompleteness
- measurePersonCompleteness
- isStandardValidConcept
- standardConceptRecordCompleteness
- sourceConceptRecordCompleteness
- sourceValueCompleteness
- plausibleValueHigh
- plausibleValueLow
- withinVisitDates


vignettes/checks/sourceConceptRecordCompleteness.Rmd
```diff
@@ -23,24 +23,50 @@ The number and percent of records with a value of 0 in the source concept field
 
 ## Definition
 
-- *Numerator*: 
-- *Denominator*:
-- *Related CDM Convention(s)*:
-- *CDM Fields/Tables*:
-- *Default Threshold Value*:
+- *Numerator*: The number of rows with a value of 0 in the `X_source_concept_id` source concept field.  In the case of `MEASUREMENT.unit_source_concept_id` and `OBSERVATION.unit_source_concept_id`, the number of rows with a value of 0 in the `X_source_concept_id` source concept field AND a non-NULL `value_as_number`. 
+- *Denominator*: The total number of rows in the table. In the case of `MEASUREMENT.unit_source_concept_id` and `OBSERVATION.unit_source_concept_id`, the number of rows with non-NULL `value_as_number`.
```
MaximMoinat, line 27: It might be because it is late, but I can't find `OBSERVATION.unit_source_concept_id`. Only Measurement and Device Exposure have a unit_source_concept_id.

vignettes/checks/sourceConceptRecordCompleteness.Rmd
```diff
@@ -23,24 +23,50 @@ The number and percent of records with a value of 0 in the source concept field
...
+```sql
+SELECT  
+  concept.concept_name AS standard_concept_name, 
+  cdmTable.X_concept_id, -- standard concept ID field for the table 
+  c2.concept_name AS source_value_concept_name, 
+  cdmTable.X_source_value, -- source value field for the table 
+  COUNT(*) 
+FROM @cdmDatabaseSchema.@cdmTableName cdmTable 
+LEFT JOIN @vocabDatabaseSchema.concept ON concept.concept_id = cdmTable.X_concept_id 
+-- WARNING this join may cause fanning if a source value exists in multiple vocabularies
```
MaximMoinat, line 54: Had to look up 'fanning', but seems appropriate :D

vignettes/checks/sourceConceptRecordCompleteness.Rmd
```diff
@@ -23,24 +23,50 @@ The number and percent of records with a value of 0 in the source concept field
...
+If source values return legitimate matches on concept_code, it’s possible that there is an error in the concept mapping step of your ETL.  Please note that while the `X_source_concept_id` fields are technically not required, it is highly recommended to populate them with OMOP concepts whenever possible.  This will greatly aid analysts in understanding the provenance of the data. 
 
+If source values do NOT return matches on concept_code and you are NOT handling concept mapping locally for a non-OMOP source vocabulary, then you likely have a malformed source code or one that does not exist in the OMOP vocabulary.  Please see the documentation in the [standardConceptRecordCompleteness](standardConceptRecordCompleteness.html) page for instructions on how to handle this scenario. 
 
 ### Data Users
+Since most standard OHDSI analytic workflows rely on the standard concept field and not the source concept field, failures of this check will not necessarily impact your analysis.  However, if your analysis depends on understanding source coding practices or on codes you know may not be fully mapped to OMOP standard concepts, then this will be a critical check failure to understand.
```
MaximMoinat, line 70:
```suggestion
Since most standard OHDSI analytic workflows rely on the standard concept field and not the source concept field, failures of this check will not necessarily impact your analysis.  However, having the source concept will give you a better understanding of the provenance of the code and highlight potential issues where meaning is lost due to mapping to a standard concept.
```

MaximMoinat (commented):
Very complete. Could not help myself to make a number of suggestions. Did not yet review the last four (sourceValueCompleteness, plausibleLow/High and withinVisitDates).

Also, will add a commit to add % for all thresholds, to be consistent with the existing documentation pages. There are some other consistency things as well we might want to look at eventually (use of ` `` ` or table/field names, capitalisation, use of `@cdmTable`/`@cdmField`, query formatting, field vs column, record vs row).

Lastly want to note that, while reading through, the descriptions quickly become quite abstract. Examples might help, something for a next version.

vignettes/checks/measurePersonCompleteness.Rmd
```diff
@@ -14,33 +14,60 @@ output:
...
 ```
 
 
 ### ETL Developers
 
+#### Observation period
+All persons in the CDM must have an observation period; OHDSI analytics tools only operate on persons with observable time, as represented by one or more observation periods.  Persons missing observation periods may represent a bug in the ETL code which generates observation periods. Alternatively, some persons may have no observable time in the source data. These persons should be removed from the CDM. 
+
+#### All other tables
+Action on persons missing records in other clinical event tables will depend on the characteristics of the source database.  In certain cases, missingness is expected – some persons may just not have a given type of data available in the source.  In others, various ETL issues may result in persons missing records in a given event table:
```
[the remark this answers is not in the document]
katy-sadowski: i like adding the suggestion, but what about measurements and procedures in an EHR database?  maybe could give an even more specific example of a reason why a source would not have a certain type of data?

vignettes/checks/sourceConceptRecordCompleteness.Rmd
katy-sadowski: whoops, that was my error - will update to reference device instead of observation here

vignettes/checks/measurePersonCompleteness.Rmd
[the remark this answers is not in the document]
MaximMoinat: Agree, it is more complex than I put it there. We can make this very generic:
> For instance, **in most OMOP database**, one would expect most persons have at least a record in the `visit_occurrence`, `condition_occurrence` and `drug_exposure` tables.

And remove the last sentence.

vignettes/checks/sourceConceptRecordCompleteness.Rmd
MaximMoinat: I see what happened, we use the same query for both standard and source concept id, with the exception for unit concepts.
https://github.com/OHDSI/DataQualityDashboard/blob/6ef7ee2dd1116741e3fe9907ef4d9cc98eccb96c/inst/sql/sql_server/field_concept_record_completeness.sql#L37

Note that we don't have this exception for the device table (`unit_concept_id` nor `unit_source_concept_id`) as it does not have a `value_as_number`. I assume it refers to the `quantity` field. Same for specimen table.

So we actually need to update the query to be consistent. It becomes a bit messy, but this would be the new where-clause.
```sql
WHERE cdmTable.@cdmFieldName = 0 
{@cdmTableName IN ('MEASUREMENT', 'OBSERVATION') & @cdmFieldName IN ('UNIT_CONCEPT_ID', 'UNIT_SOURCE_CONCEPT_ID')}?{AND cdmTable.value_as_number IS NOT NULL}
{@cdmTableName IN ('DEVICE_EXPOSURE', 'SPECIMEN') & @cdmFieldName IN ('UNIT_CONCEPT_ID', 'UNIT_SOURCE_CONCEPT_ID')}?{AND cdmTable.quantity IS NOT NULL}
```
That said, after thinking this through for a bit, if the ETL is properly defined it should not be an issue (i.e. if value is empty, then also unit should be empty)

katy-sadowski: Oh my!  Great catch; we don't even check this on source concept ID fields, or any field in device or specimen.  So for now I will update this to remove the language altogether for source concept ID.  I'll file an issue to get this updated.

vignettes/checks/isStandardValidConcept.Rmd
```diff
@@ -23,24 +23,55 @@ The number and percent of records that do not have a standard, valid concept in
...
+SELECT  
+  '@cdmTableName.@cdmFieldName' AS violating_field,  
+  cdmTable.*  
+FROM @schema.@cdmTableName cdmTable 
+  JOIN @vocabDatabaseSchema.concept co ON cdmTable.@cdmFieldName = co.concept_id 
+WHERE co.concept_id != 0  
+  AND (co.standard_concept != 'S' OR co.invalid_reason IS NOT NULL) 
 ```
 
+You may build upon this query by joining the relevant `X_concept_id` and `X_source_concept_id` columns to the concept table and inspecting their names and vocabularies.  If the `X_source_concept_id` correctly represents the source code in `X_source_value`, the fix will be a matter of ensuring your ETL is correctly using the concept_relationship table to map the source concept ID to a standard concept via the ‘Maps to’ relationship.  If you are not populating the `X_source_concept_id` column and/or are using an intermediate concept mapping table, you may need to inspect the mappings in your mapper table to ensure they’ve been generated correctly using the ‘Maps to’ relationship for your CDM’s vocabulary version.
```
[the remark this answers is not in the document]
katy-sadowski: updated the language to make it more clear the source concept ID is being used to investigate the mapping issue.

katy-sadowski:
Thank you SO much for your review @MaximMoinat !  Your feedback and suggestions were extremely useful.  I will plan to merge this in tomorrow, in case you want to take a final look.  Then, I will merge in your other doc PR (https://github.com/OHDSI/DataQualityDashboard/pull/557), and then I will do the release!

vignettes/checks/sourceConceptRecordCompleteness.Rmd
MaximMoinat: @katy-sadowski Opened the issue here: https://github.com/OHDSI/DataQualityDashboard/issues/558

katy-sadowski: Thank you!!!

PR #574, opened 2024-10-24
Fix missing pull() and wrong launchLogViewer() file.path() argument fix #568 

### LINKED ISSUE #568: Missing `pull()` in CodeToRun exclude checks workflow
https://github.com/OHDSI/DataQualityDashboard/blob/98c825a9449a5d15413669a7e931f98b1e1c51c9/extras/codeToRun.R#L95

Also, we should call library(dplyr) at the top.

Also, this shouldn't have cdmSourceName in it: https://github.com/OHDSI/DataQualityDashboard/blob/98c825a9449a5d15413669a7e931f98b1e1c51c9/extras/codeToRun.R#L121

samhillman:
fix #568 . Update the checksToExclude by changing the select() to a pull() to make sure we return the vector only rather than a column. Add explicit namespacing and call the library(dplyr) at the top of the commented code. Update the inspect logs code to correct the file.path() argument to include the cdmSourceName only in the .txt file name.

### DIFF
--- a/extras/codeToRun.R
+++ b/extras/codeToRun.R
@@ -88,11 +88,12 @@ checkNames <- c() # Names can be found in inst/csv/OMOP_CDM_v5.3_Check_Descripti
 
 # want to EXCLUDE a pre-specified list of checks? run the following code:
 #
+# library(dplyr)
 # checksToExclude <- c() # Names of check types to exclude from your DQD run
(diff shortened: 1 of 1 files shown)


katy-sadowski (approved):
Woohoo! Thanks!

Issue #538, opened 2024-04-11
Add executeDqChecks parameter for check type severity


katy-sadowski:
We have added a new column to the check descriptions csvs for check type severity (`severity`).  Similar to how users can specify which check levels to run DQD for (https://github.com/OHDSI/DataQualityDashboard/blob/6ef7ee2dd1116741e3fe9907ef4d9cc98eccb96c/R/executeDqChecks.R#L37), we should add the ability to specify and run DQD for chosen severity level(s).

Not all checks have a severity populated in the check descriptions files, but these will be completed in time for the next release and can be coordinated with the addition of this new parameter.

PR #576, opened 2024-10-27
Added parameter for checkSeverity

ob325:
hades hackathon enhancement for #538


katy-sadowski:
@ob325 thank you so much for this contribution!  It was wonderful meeting you at the hackathon and I really appreciate you following through on this PR 😄 

I think the changes all look good.  There are just 2 things left to do before we can merge this in:

- R CMD Check is throwing a warning because the docs need to be regenerated.  If you run `devtools::document()` in the R console within the project, that should fix the warning
- It'd also be great to update our vignette with info about the new parameter.  I'm thinking you could just add a couple lines here showing how to set the checkSeverity parameter, and add it to the executeDqChecks function call: https://github.com/OHDSI/DataQualityDashboard/blob/98c825a9449a5d15413669a7e931f98b1e1c51c9/vignettes/DataQualityDashboard.rmd#L112-L133

Let me know if those additions sound doable to you!  Thanks again!

Issue #611, opened 2025-06-16
Add ng/mL and /μL units to Measurement Plausibility rules


dearzion:
### Request to Add Additional Units to Measurement Plausibility Rules

**1. TESTOSTERONE [MASS/VOLUME] IN SERUM OR PLASMA**
- Concept ID: 3008893  
- Request: Add `unit_concept_id = 8842` (ng/mL)  
- Justification: "ng/mL" is a commonly used unit for testosterone tests in clinical laboratories. This unit exists in the OMOP vocabulary and should be included in the list of plausible units.

**2. NEUTROPHILS [#/VOLUME] IN BLOOD BY AUTOMATED COUNT**  
**3. EOSINOPHILS [#/VOLUME] IN BLOOD BY AUTOMATED COUNT**
- Concept IDs: 3013650, 3028615  
- Request: Add `unit_concept_id = 8647` (/μL)  
- Justification: Some lab results report WBC subtypes using simple counts per microliter (e.g., 200/μL). The unit `/μL` (unit_concept_id: 8647) is used in real-world clinical data and is already defined in the OMOP vocabulary. Including this unit will enhance DQD's ability to reflect actual practice.

Issue #558, opened 2024-07-12
Make consistent unit concept exception in Concept Record Completeness


MaximMoinat:
This issue came up while reviewing the documentation of sourceConceptRecordCompleteness check. 

In the query for both standard and source concept id, we have this exception for unit concepts:
https://github.com/OHDSI/DataQualityDashboard/blob/6ef7ee2dd1116741e3fe9907ef4d9cc98eccb96c/inst/sql/sql_server/field_concept_record_completeness.sql#L37

The reason we have this rule is that many databases will enter a 0 as the `unit_concept_id` by default, even if there is no measurement/observation value. These should be ignored to get a meaningful violating percentage. (fyi: the correct way to do this is to leave it empty, NULL, as the unit concept is not a required field).

However, the rule is inconsistent; we don't apply the exception to the `unit_source_concept_id` and not to the device and specimen table which also have a `unit_concept_id` and `unit_source_concept_id`. For these tables, we can assume it refers to the `quantity` field (instead of value_as_number).

We need to update the query to be consistent. It becomes a bit messy, but this would be the new where-clause.
```sql
WHERE cdmTable.@cdmFieldName = 0 
{@cdmTableName IN ('MEASUREMENT', 'OBSERVATION') & @cdmFieldName IN ('UNIT_CONCEPT_ID', 'UNIT_SOURCE_CONCEPT_ID')}?{AND cdmTable.value_as_number IS NOT NULL}
{@cdmTableName IN ('DEVICE_EXPOSURE', 'SPECIMEN') & @cdmFieldName IN ('UNIT_CONCEPT_ID', 'UNIT_SOURCE_CONCEPT_ID')}?{AND cdmTable.quantity IS NOT NULL}
```

katy-sadowski:
Thanks so much for the thorough summary @MaximMoinat !  I agree with your assessment.  I think we might also want to consider adding a check for cases B, D, and E.  I observe inconsistencies in these fields often and it's confusing.  We should be enforcing that non-required concept fields are left NULL when there is no data with which to populate a concept.  This way users can distinguish no data from unmappable data.

Issue #510, opened 2023-11-24
New check request: overlapping observation periods


MaximMoinat:
From the OMOP Observation Period conventions:
> For each Person, one or more OBSERVATION_PERIOD records may be present, but they will **not overlap** or **be back to back to each other.**

The former is checked, the latter two conditions are not. Below query would give you the violating records for overlapping observation periods. By adding one day to the end dates, this also captures observation periods ending and starting within a day (i.e. back to back, without a gap).

```sql
SELECT *
FROM observation_period AS a
JOIN observation_period AS b ON
      a.person_id = b.person_id 
  AND a.observation_period_start_date <= DATEADD(d, 1, b.observation_period_end_date)
  AND DATEADD(d, 1, a.observation_period_end_date) >= b.observation_period_start_date
  AND a.observation_period_id <> b.observation_period_id
;
```

katy-sadowski:
Love this idea, this is a very important one to check.

Issue #493, opened 2023-10-06
Allow the situation when Condition Era is not built from Condition occurrence where condition_concept_id =0


dimshitc:
Please edit the measureConditionEraCompleteness checl

The number and Percent of persons that does not have condition_era built successfully for all persons in condition_occurrence.
It should omit Condition records where condition_concept_id = 0. In this case it's OK, that era is not built

katy-sadowski:
Agree, good catch. People whose only CO rows have 0 concept_id, and people who have no conditions at all (already excluded from the check), should not be expected to have condition eras. And we'll flag the missed standard concept mapping in another check.

MaximMoinat:
Note that the default condition era derivation script does NOT filter for condition_concept_id = 0:
https://ohdsi.github.io/CommonDataModel/sqlScripts.html#Condition_Eras

PR #618, opened 2025-08-04
Check logic updates and new check

katy-sadowski:
#510 - adds check for overlapping and b2b observation periods

#611 - adds some missing plausible units

#493 - excludes persons without non-zero condition_occurrence rows from measureConditionEraCompleteness

#558 - removes value as number exception from measureConceptCompleteness; adds null values to numerator for non-required concept ID fields; limits denominator to non-null concept ID or non-null source value for non-required fields.  These changes will discourage use of 0 as placeholder for missing units/statuses/etc. - non-required concept fields should be left null if there is no source value to map.  The changes will also flag cases where a source value is present but no concept ID is populated, which are currently not covered in DQD.

Related to the above, this PR also adds a non-NULL requirement to the numerator and denominator of isStandardValidConcept.  Columns with a missing standard concept will be flagged either in isRequired or measureConceptCompleteness, depending on whether the field is required.


inst/sql/sql_server/field_is_standard_valid_concept.sql
```diff
@@ -54,6 +54,7 @@ FROM
   		JOIN @cohortDatabaseSchema.@cohortTableName c 
   		    ON cdmTable.person_id = c.subject_id
   		    AND c.cohort_definition_id = @cohortDefinitionId
-    }
+    	}
+	WHERE cdmTable.@cdmFieldName IS NOT NULL
```
katy-sadowski, line 58: this change removes overlap between this check and isRequired / standardConceptRecordCompleteness

inst/sql/sql_server/concept_plausible_unit_concept_ids.sql
```diff
@@ -39,8 +39,11 @@ FROM
     		AND c.cohort_definition_id = @cohortDefinitionId
     	}
 		WHERE m.@cdmFieldName = @conceptId
-		  	AND COALESCE (m.unit_concept_id, -1) NOT IN (@plausibleUnitConceptIds) -- '-1' stands for the cases when unit_concept_id is null
-		  	AND m.value_as_number IS NOT NULL 
+		  	AND /* '-1' stands for the cases when the only plausible unit_concept_id is no unit; 0 prevents flagging rows with a unit_concept_id of 0, which are checked in standardConceptRecordCompleteness */
+				CASE 
+					WHEN '@plausibleUnitConceptIds' = '-1' THEN COALESCE(m.unit_concept_id, -1) NOT IN (@plausibleUnitConceptIds) 
+					ELSE m.unit_concept_id NOT IN (@plausibleUnitConceptIds, 0) 
+				END
```
MaximMoinat, line 46: We can simplify this to "fails if plausible -1 but unit not null, or unit not in plausible units".
Also, instead of letting 0 always pass, we should simply ignore it for the check (see next comment).
```suggestion
 (
 (@plausibleUnitConceptIds = -1 AND m.unit_concept_id IS NOT NULL)
 OR m.unit_concept_id NOT IN (@plausibleUnitConceptIds) 
 )
```

inst/sql/sql_server/concept_plausible_unit_concept_ids.sql
```diff
@@ -54,6 +57,5 @@ FROM
     		AND c.cohort_definition_id = @cohortDefinitionId
   	}
 	WHERE m.@cdmFieldName = @conceptId
-	  	AND value_as_number IS NOT NULL
```
MaximMoinat, line 57: AND unit_concept_id != 0 to exclude 0 from denominator (which are checked in standardConceptRecordCompleteness)

inst/sql/sql_server/field_concept_record_completeness.sql
```diff
@@ -34,7 +35,27 @@ FROM (
                 ON cdmTable.person_id = c.subject_id
                 AND c.cohort_definition_id = @cohortDefinitionId
         }
-        WHERE cdmTable.@cdmFieldName = 0 {@cdmFieldName == 'UNIT_CONCEPT_ID' & (@cdmTableName == 'MEASUREMENT' | @cdmTableName == 'OBSERVATION')}?{AND cdmTable.value_as_number IS NOT NULL}
+        WHERE cdmTable.@cdmFieldName = 0
+        {@cdmTableName != 'DOSE_ERA' & (@cdmFieldName == 'UNIT_CONCEPT_ID' | @cdmFieldName == 'UNIT_SOURCE_CONCEPT_ID')}?{OR (cdmTable.@cdmFieldName IS NULL AND cdmTable.unit_source_value IS NOT NULL)}
```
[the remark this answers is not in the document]
MaximMoinat: Indeed hideous, and took me a few minutes to understand. It might warrant making a separate check for the non-required concept_id fields (where we do give it a source value parameter).
For now, I propose the following comment above this where statement for future developers to explain:
```SQL
/* Violates if 0, or if empty and respective source value non-empty. For example, this resolves to the following for death.cause_concept_id:
WHERE death.cause_concept_id = 0 OR (death.cause_concept_id IS NULL AND cdmTable.cause_source_value IS NOT NULL)
*/

File: inst/sql/sql_server/table_observation_period_overlap.sql [hunk 5]
```diff
@@ -0,0 +1,59 @@
...
+    		    ON cdmTable.person_id = c.subject_id
+    		    AND c.cohort_definition_id = @cohortDefinitionId
+    	}
+		JOIN @cdmDatabaseSchema.observation_period cdmTable2 
+		    ON cdmTable.person_id = cdmTable2.person_id
+		    AND cdmTable.observation_period_id != cdmTable2.observation_period_id
+		WHERE (cdmTable.observation_period_start_date <= cdmTable2.observation_period_end_date 
+		    AND cdmTable.observation_period_end_date >= cdmTable2.observation_period_start_date)
+		    OR (DATEADD(day, 1, cdmTable.observation_period_end_date) = cdmTable2.observation_period_start_date)
+		    OR (DATEADD(day, 1, cdmTable2.observation_period_end_date) = cdmTable.observation_period_start_date)
```
MaximMoinat, line 45: Note that this counts every pair of overlapping observation periods twice.

DESCRIPTION
```diff
@@ -42,7 +42,7 @@ Suggests:
     markdown,
     shiny,
     ggplot2,
-    Eunomia,
+    Eunomia (>= 2.0.0),
```
katy-sadowski, line 45: needed due to duckdb support

inst/sql/sql_server/concept_plausible_unit_concept_ids.sql
MaximMoinat: I think you made a very good argument for the change :-)

PR #661, opened 2026-02-17
Feature/pipeline mode

james-cockayne:
Hi,

Here's an early version of the kind of change that I was thinking.

In this PR I've added a few new things

New features
- Option to log errors to the error console
- Option to exclude specific tests (used later to exclude justifiable failures)

I've also added a couple of other bits
- A docker definition that can encapsulate the DQD package, including all of its dependencies.
- An example `main.r` script that could be used as an entry point for the docker image
- A VS Code dev container definition, used for development purposes only

What's missing
- Some decent documentation and explanation of how it works and how to use it
- More thorough testing. I will implement it in both of our large pipelines (around 19m patients) 

I'm adding this PR now just because of the time difference, and I'm finishing work for the day.

Let me know what you think


R/executeDqChecks.R
```diff
@@ -120,7 +123,8 @@ executeDqChecks <- function(connectionDetails,
   stopifnot(
     is.null(checkNames) | is.character(checkNames),
     is.character(checkSeverity),
-    is.null(tablesToExclude) | is.character(tablesToExclude)
+    is.null(tablesToExclude) | is.character(tablesToExclude),
+    is.null(checksToExclude) | is.character(checksToExclude)
```
katy-sadowski, line 127: let's also stop if they fill out both checksToExclude and checkNames

and now i'm typing this, maybe it should be checkNamesToExclude for consistency?

james-cockayne:
Moved this logic to https://github.com/james-cockayne/dqd_console_results

PR #682, opened 2026-08-14
Execute DQ on Extension with Long format thresholds

MaximMoinat:
~Adds function `executeDqChecksOnExtension` which~ Updated `executeDqChecks` to take check thresholds in long format and only executes on these tables/fields. See vignette for explanation and minimal usage example.

~Name of the function to be decided.~ The defining feature is that you can specify the thresholds in a long format, which makes it easier to define for custom/extension tables.

While this is a stand-alone function that uses the DQD machinery, the following small changes to DQD were needed. Behaviour should have remained the same.
 - `isForeignKey.sql` refactored to allow checking against non-CDM tables. Foreign key tables were hardcoded.
 - `runCheck.R` refactoring to handle when check not found in threshold (in same logic as no rows found with enabled check)
 - `testthat/setup.R` moved test threshold files to the testthat folder and read these with `test_path()`


extras/thresholdWideToLong.R
```diff
@@ -0,0 +1,306 @@
...
+    cdmTableName,
+    !c(
+      conceptPrefix,
+      tableDescription,
+      validation,
+      userGuidance,
+      etlConventions,
+      ends_with("Threshold"),
+      ends_with("Notes")
+    )
```
katy-sadowski, line 32: Instead could you filter to the column names matching the check names in the check descriptions files? Would make it more robust to new cols being added.

extras/thresholdWideToLong.R
```diff
@@ -0,0 +1,306 @@
...
+    by = join_by(cdmTableName, schema, checkName)
+  ) %>%
+  mutate(
+    checkValue = if_else(checkValue == 'Yes', NA, checkValue)
+  ) %>%
+  select(
+    cdmTableName,
+    checkName,
+    Threshold,
+    checkParameter = checkValue,
```
katy-sadowski: what is this for?

R/executeDqChecksOnExtension.R
```diff
@@ -0,0 +1,320 @@
...
+#' @param cohortDatabaseSchema      The schema where the cohort table is located.
+#' @param cohortTableName           The name of the cohort table. Defaults to `cohort`.
+#' @param tablesToExclude           (OPTIONAL) Choose which CDM tables to exclude from the execution.
+#' @param cdmVersion                The CDM version to target for the data source. Options are "5.2", "5.3", or "5.4". By default, "5.3" is used.
+#'
+#' @importFrom DataQualityDashboard executeDqChecks
+#' @importFrom utils packageVersion getFromNamespace
+#' @return dqd results object
+#' @export
+executeDqChecksOnExtension <- function(
```
katy-sadowski: what's the reason for making a separate function, vs optionally supporting a long format input to `executeDqChecks`?

katy-sadowski (commented):
Thanks so much for this @MaximMoinat !!  A couple of questions below.

I fully support the long format option.  I agree that maybe the naming could be more generic, because I could see this being useful for other use cases as well (but would need to brainstorm a bit).

extras/thresholdWideToLong.R
MaximMoinat: To only keep relevant columns. Will merge with the `mutate`

extras/thresholdWideToLong.R
MaximMoinat: I see your point. It is however not that straightforward as some checks are named slightly different in the thresholdsfile (e.g. `isForeignKey` vs `fkTableName`). Also, it is just a sample function that works on current thresholds.

R/executeDqChecksOnExtension.R
MaximMoinat: I did not want to overflow the executeDQD with more parameters, and make this an explicit choice. That said, we could build it such that `executeDqChecks` detects if given thresholds are in long-format and does the conversion in the background. I will give it a shot.

MaximMoinat:
@katy-sadowski Updated `executeDqChecks` to take long-format threshold files. This is indeed a bit cleaner. Detection of long-format is based on the column name 'checkName' appearing in the given thresholds file.
Removed the new `executeDqChecksOnExtension` function.

Issue #314, opened 2022-04-15
isTemporallyConstant and validPrevalence thresholds without check description


MaximMoinat:
Looking at the [Concept_Level threshold file](https://github.com/OHDSI/DataQualityDashboard/blob/main/inst/csv/OMOP_CDMv5.4_Concept_Level.csv), I noticed that there are two checks that are not part of the check descriptions: `isTemporallyConstant` and `validPrevalence`. From searching in previous threads (#46),  I understood that the `validPrevalence` was not yet included because this is non-trivial and we first have to build a knowledge base. I assume same is true for `isTemporallyConstant`

However, let's say I really want to run these two check types (maybe with custom db-specific prevalence ranges), is there already code available to do so?

(sorry for bringing this hidden feature up!)

clairblacketer:
Hi @MaximMoinat unfortunately no, as far as I am aware there is no code available for these checks.

clairblacketer:
@MaximMoinat since these checks are just "nice to haves" but don't have code to support, can we close this issue? I would like to explore the idea of a benchmarking study for DQD and characterization after the symposium where we could come back aroudn to the idea of validPrevalence

MaximMoinat:
Works for me. My only concern is that the thresholds file has columns for a check that is not implemented, confusing people (like me). But if we are coming back to this later, that should be fine.

PR #686, opened 2026-09-04
Add CDM v5.5 support and consolidate concept level thresholds

katy-sadowski:
### Rationale
Add support for OMOP CDM v5.5, and clean up two sources of drift in the threshold files found along the way.

### Changes
- **CDM v5.5 support.** New `OMOP_CDMv5.5_{Table,Field}_Level.csv` and `Check_Descriptions.csv`; `cdmVersion = "5.5"` now accepted. The v5.5 files are a superset of v5.4, covering the 24 fields and 3 tables v5.5 adds. The new vocabulary tables (`PACK_CONTENT`, `CONCEPT_METADATA`, `CONCEPT_RELATIONSHIP_METADATA`) are added to the default `tablesToExclude`.
- **One concept level file.** The v5.2/v5.3/v5.4 concept files were meant to be identical but had drifted; they are replaced by `OMOP_CDM_Concept_Level.csv`.
- **Refreshed CDM docs columns.** `userGuidance`, `etlConventions`, and `tableDescription` in the v5.3/v5.4/v5.5 files now match the current CommonDataModel spec. Documentation only.

### Notes
- Concept file consolidation changes results slightly: v5.4 unchanged, v5.2 gains 9 cholesterol/creatinine checks dropped in error, and v5.3 loses two implausible unit concepts on Sodium (cell count units on a molar concentration measurement).
- `OBSERVATION.value_as_date` holds a value rather than an event date, so temporal plausibility checks are left off to avoid false positives.
- v5.2 doc columns are untouched, as CDM v5.2 is no longer published in the CommonDataModel repo.
- New tests cover threshold file loading only. Eunomia is a v5.3 database, so field level checks cannot run cleanly against it at v5.4 or v5.5.
- `DESCRIPTION` is not bumped and `docs/` is not regenerated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012jnm3W2EfddVLW6d5e5yNs


man/executeDqChecks.Rd
```diff
@@ -28,7 +28,8 @@ executeDqChecks(
   cohortDatabaseSchema = resultsDatabaseSchema,
   cohortTableName = "cohort",
   tablesToExclude = c("CONCEPT", "VOCABULARY", "CONCEPT_ANCESTOR",
-    "CONCEPT_RELATIONSHIP", "CONCEPT_CLASS", "CONCEPT_SYNONYM", "RELATIONSHIP", "DOMAIN"),
+    "CONCEPT_RELATIONSHIP", "CONCEPT_CLASS", "CONCEPT_SYNONYM", "RELATIONSHIP", "DOMAIN",
```
MaximMoinat, line 31: Drug strength was missing here.
```suggestion
    "CONCEPT_RELATIONSHIP", "CONCEPT_CLASS", "CONCEPT_SYNONYM", "RELATIONSHIP", "DOMAIN", "DRUG_STRENGTH",
```

R/executeDqChecks.R
```diff
@@ -82,7 +83,10 @@ executeDqChecks <- function(connectionDetails,
                             cohortDefinitionId = c(),
                             cohortDatabaseSchema = resultsDatabaseSchema,
                             cohortTableName = "cohort",
-                            tablesToExclude = c("CONCEPT", "VOCABULARY", "CONCEPT_ANCESTOR", "CONCEPT_RELATIONSHIP", "CONCEPT_CLASS", "CONCEPT_SYNONYM", "RELATIONSHIP", "DOMAIN"),
+                            tablesToExclude = c(
+                              "CONCEPT", "VOCABULARY", "CONCEPT_ANCESTOR", "CONCEPT_RELATIONSHIP", "CONCEPT_CLASS", "CONCEPT_SYNONYM", "RELATIONSHIP", "DOMAIN",
```
MaximMoinat, line 87:
```suggestion
                              "CONCEPT", "VOCABULARY", "CONCEPT_ANCESTOR", "CONCEPT_RELATIONSHIP", "CONCEPT_CLASS", "CONCEPT_SYNONYM", "RELATIONSHIP", "DOMAIN", "DRUG_STRENGTH",
```

vignettes/Thresholds.rmd
```diff
@@ -27,15 +27,15 @@ There is a set of three csv files that underlie the DQD. These files indicate wh
...
 
 ![Figure 1: Field Level Control File Example](images/fieldLevelFile.png)
 
-- **OMOP_CDMvx.x.x_Concept_Level.csv**: This file has a list of all concepts and any checks to be performed on that concept. For example, there are checks that evaluate biologically plausible ranges for measurement values, like [plausibleValueLow](https://ohdsi.github.io/DataQualityDashboard/articles/CheckTypeDescriptions.html#plausiblevaluelow-1). If you would like to change the failure thresholds for any of these checks copy this file to a location the R instance can access.
+- **OMOP_CDM_Concept_Level.csv**: This file has a list of all concepts and any checks to be performed on that concept. Concept level thresholds do not vary by CDM version, so a single file is shared across all supported versions. For example, there are checks that evaluate biologically plausible ranges for measurement values, like [plausibleValueLow](https://ohdsi.github.io/DataQualityDashboard/articles/CheckTypeDescriptions.html#plausiblevaluelow-1). If you would like to change the failure thresholds for any of these checks copy this file to a location the R instance can access.
```
MaximMoinat, line 38: The Concept-level plausibleValueLow is not executed anymore. Propose to update the example to:

```suggestion
- **OMOP_CDM_Concept_Level.csv**: This file has a list of all concepts and any checks to be performed on that concept. Concept level thresholds do not vary by CDM version, so a single file is shared across all supported versions. For example, there are checks that evaluate plausible units for measurements and plausible gender for conditions. If you would like to change the failure thresholds for any of these checks copy this file to a location the R instance can access.
```

NEWS.md
```diff
@@ -1,3 +1,29 @@
...
+
+### Concept level threshold files
+
+**The v5.2, v5.3, and v5.4 concept level threshold files have been replaced by a single file, `OMOP_CDM_Concept_Level.csv`.** Concept level thresholds do not vary by CDM version, and the three files had drifted apart unintentionally. If you pass `conceptCheckThresholdLoc` you are unaffected; if you rely on the default file, note the following changes:
+
+- v5.4 results are unchanged; the consolidated file is identical to the previous v5.4 file
+- v5.2 gains 9 `OBSERVATION.OBSERVATION_CONCEPT_ID` cholesterol and creatinine `plausibleValueLow`/`plausibleValueHigh` checks, which were present for v5.3 and v5.4 but had been dropped from the v5.2 file in error
```
MaximMoinat, line 24: I propose we remove these 'added' checks om the consolidated file, as the plausibleValueLow/plausibleValueHigh are not executed anymore.

MaximMoinat: Same for change in plausibleUnitConceptId, we could remove it entirely.

MaximMoinat (commented):
Tested on my CDM v5.5 instance and result looks like expected. In total 89 additional check results for the new tables/fields and no missing results. The merged concept level will help in future maintenance.

Before approving, I have two suggestions:
 - Exclude the vocabulary table `DRUG_STRENGTH` by default.
 - The concept level file lists thresholds for checks that are not executed. Would be good to use the merge to also clean this file. Will make stacked PR for that.

NEWS.md
katy-sadowski: Agree!  Can you please update NEWS in your PR #688 ?

katy-sadowski:
@MaximMoinat thanks for testing!

- Re: DRUG_STRENGTH, I (Claude) noticed that too, but I didn't add it because it'd technically be a breaking change for users who assume it's in their DQD output.
- Agree re: removing those concept checks. You reported this in #314 a long time ago, at which time I think there was still interest in adding them, but I think we can safely remove them now.

MaximMoinat:
@katy-sadowski Still want to make a case for removing `DRUG_STRENGTH` by default. It results in a failing check because of some ingredient concepts being from CVX class instead of ingredient, which is a 'false positive'.

PR #691, opened 2026-09-10
Consolidate check descriptions

MaximMoinat:
Consolidates the separate check description files, which were identical across cdm versions. This is a similar fix to the concept level threshold consolidation. Leaving only field and table level thresholds per cdm version.

Updates the news.md, reflect also the concept level cleaning.


katy-sadowski (approved):
Good call! Thanks!

katy-sadowski (approved):

